PR #22, opened 2019-05-14
Update metadata api to use from data

jonathanthiry:


src/main/scala/se/lu/nateko/cp/meta/routes/LinkedDataRoute.scala
```diff
@@ -105,6 +99,17 @@ object LinkedDataRoute {
 					complete(MetaDb.getDobjGraphInfos(config))
 				}
 			}
+		} ~
+		options{
+			pathPrefix("objects" / Sha256Segment) { _ =>
+				respondWithHeaders(
+					`Access-Control-Allow-Origin`.*,
+					`Access-Control-Allow-Methods`(HttpMethods.GET, HttpMethods.POST),
```
mirzov: Only HTTP GET should be enough here

PR #25, opened 2019-08-12
Refactor upload gui

jonathanthiry:


uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FormComponents.scala
```diff
@@ -169,52 +187,11 @@ class DoiOptInput(elemId: String, cb: () => Unit) extends GenericOptionalInput[D
...
-	s => if (s == null || s.trim.isEmpty) Success(None) else parser(s.trim)
-)
-
-class UriListInput(elemId: String, cb: () => Unit) extends GenericTextArea[Seq[URI]](elemId, cb, fail("Missing list of object urls"))(s =>
+class UriListInput(elemId: String, cb: () => Unit) extends GenericTextInput[Seq[URI]](elemId, cb, fail("Missing list of object urls"))(s =>
 	Try(s.split("\n").map(new URI(_)))
 )
 
-class OptTextArea(elemId: String, cb: () => Unit) extends GenericOptionalTextArea[String](elemId, cb)(s => Try(Some(s)))
+class OptTextArea(elemId: String, cb: () => Unit) extends GenericOptionalInput[String](elemId, cb)(s => Try(Some(s)))
```
mirzov: It seems this class is redundant, as TextOptInput can be used instead

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/Form.scala
```diff
@@ -15,46 +15,40 @@ class Form(
...
-			for(dto <- dto; file <- fileInput.file) {
-				onUpload(dto, Some(file))
-			}
-		case Some("collection") =>
+		case Collection =>
 			for(dto <- dto) {
 				onUpload(dto, None)
 			}
-		case _ => ()
+		case _ =>
```
mirzov: Better not to use catch-all here and in other similar places, but to instead rely on compiler's pattern-match-exhaustion check.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FormComponents.scala
```diff
@@ -99,6 +104,14 @@ class Radio(elemId: String, cb: String => Unit) {
 	}
 }
 
+class FormTypeRadio(elemId: String, cb: String => Unit) extends Radio(elemId, cb) {
```
mirzov: `cb` constructor argument should have type `FormType => Unit`, not `String => Unit`. `onFileTypeSelected` in Form.scala should be changed accordingly, and not access `typeControl`.

PR #26, opened 2019-08-30
Add production metadata

jonathanthiry:


uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/Form.scala
```diff
@@ -126,7 +165,8 @@ class Form(
 		acqInterval <- timeIntevalInput.value.withErrorContext("Acqusition time interval");
 		nRows <- nRowsInput.value.withErrorContext("Number of rows");
 		samplingHeight <- samplingHeightInput.value.withErrorContext("Sampling height");
-		instrumentUri <- instrUriInput.value.withErrorContext("Instrument URI")
+		instrumentUri <- instrUriInput.value.withErrorContext("Instrument URI");
+		production <- if(productionElements.areEnabled) dataProductionDto else Try(None)
```
mirzov: Success(None) is more exact here (instead of Try(None))

src/main/twirl/views/UploadGuiPage.scala.html
```diff
@@ -62,119 +76,164 @@
...
+				</section>
+				<section class="col-md-4 data-section">
+					<div class="form-section panel panel-default">
+						<h2>Acquisition</h2>
+						<div class="form-group">
+							<label for="stationselect">Station</label>
+							<select id="stationselect" class="form-element" disabled></select>
+						</div>
+						<div class="form-group">
+							<label for="acqstartinput">From (@timezone)</label>
```
mirzov: This is probably a mistake. Time is expected to be provided in a format that includes time zone information, such as ISO-8601 with "Z" in the end.

PR #52, opened 2020-02-20
Add support for metadata updates

jonathanthiry:


mirzov:
Some desirable further improvements for feature completeness:

- File name should be editable for data/document object metadata updates
- The "Item type" radio-button should be disabled when in "update metadata" mode
- Instrument uri is also an `OptionalOneOrSeq` nowadays (currently fetching an ATC data object metadata results in string like "Left(http://meta.icos-cp.eu/resources/instruments/ATC_489)"
- previous version should probably better be base64url-encoded, not base64-encoded. `Sha256Sum.id` gives the url-encoded  version.
- switching from "Update metadata" to "New file" mode should reset the form

PR #58, opened 2020-04-16
[uploadgui] Recompute hash if file was modified

jonathanthiry:
In Firefox the file and last modification date don't change so the file correct file has to be selected again. In Chrome the last modification date changes and the hash can be recomputed.

### DIFF
--- a/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FormComponents.scala
+++ b/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FormComponents.scala
@@ -83,13 +83,13 @@ class FileInput(elemId: String, cb: () => Unit){
 	}
 
 	def hasBeenModified: Boolean =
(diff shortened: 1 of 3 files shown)


uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FormComponents.scala
```diff
@@ -68,18 +70,41 @@ class Select[T](elemId: String, labeller: T => String, autoselect: Boolean = fal
...
+	def hash: Try[Sha256Sum] = {
+		if (hasBeenModified) {
+			fail("The file has been modified, please choose the updated version")
+		} else {
+			_hash
+		}
+	}
+
+	def hasBeenModified: Boolean =
+		file.map(_.asInstanceOf[js.Dynamic].lastModified.asInstanceOf[Double]) != Success(_lastModified)
```
mirzov, line 86: Extracting `lastModified` from a `File` object is done at least twice in the code, and as it is rather "dirty" code, it should be moved to a dedicated utility function.

jonathanthiry:
Fixed in 6beb76d

PR #73, opened 2020-06-22
[station map] Fix the html/json station page cache

jonathanthiry:
Add an extra json format query so the browser doesn't get confused
between html and json format cache.


jonathanthiry:
For example:
- Go to https://meta.icos-cp.eu/resources/stations/AS_LIN
- Click on `Browse data from Lindenberg`
- Go back to the station page

The browser shows the cached json instead of the html page.

PR #78, opened 2020-08-10
Update labeling app title and description

jonathanthiry:


src/main/js/labeling/views/NavBarFactory.jsx
```diff
@@ -19,7 +19,11 @@ module.exports = function(WhoAmIStore) {
 				</div>
 				{this.props.piMode
 					? <h1>PI Information <small>for ICOS metadata</small></h1>
-					: <h1>ICOS Station Labeling <small>Steps 1, 2, 3</small></h1>}
+					: <div>
+							<h1>Labelled stations</h1>
```
mirzov, line 23: Should rather be "ICOS station labelling status", as it shows all the stations in the labelling pipeline, not only the labelled ones.

src/main/js/labeling/views/NavBarFactory.jsx
```diff
@@ -19,7 +19,11 @@ module.exports = function(WhoAmIStore) {
...
+							<div>The labelling status for ICOS stations. You can read more about our data collection process <a href="https://www.icos-cp.eu/data-services/data-collection">here</a>.</div>
```
mirzov, line 24: It seems that it would be more natural to link to either https://www.icos-cp.eu/about/join-icos/process-stations or to https://www.icos-cp.eu/data-services/data-collection/labelling-process here (by the way, the latter page has a link to the former, which is currently using a URL starting with "demo.icos-cp.eu").

src/main/twirl/views/LabelingPage.scala.html
```diff
@@ -1,4 +1,4 @@
-@CpCommonPage("ICOS station labeling service"){
+@CpCommonPage("Labelled stations - ICOS"){
```
mirzov, line 1: "Station labelling status - ICOS" ?

PR #82, opened 2020-11-02
Use envri specif theme on sparql client page

jonathanthiry:


jonathanthiry:
Next step will be to customize the list of predifined requests.

jonathanthiry:
I'm merging this now but please let me know later on if you have any comments.

PR #86, opened 2020-11-24
Improve displayed time coverage format

jonathanthiry:


src/main/scala/se/lu/nateko/cp/meta/services/citation/CitationMaker.scala
```diff
@@ -110,9 +111,17 @@ object CitationMaker{
 
 	private def getTimeFromInterval(interval: TimeInterval, zoneId: ZoneId): String = {
 		val duration = Duration.between(interval.start, interval.stop)
+		val startZonedDateTime = ZonedDateTime.ofInstant(interval.start, zoneId)
+		val stopZonedDateTime = ZonedDateTime.ofInstant(interval.stop, zoneId)
 		if (duration.getSeconds < 24 * 3601) { //daily data object
 			val middle = Instant.ofEpochMilli((interval.start.toEpochMilli + interval.stop.toEpochMilli) / 2)
 			formatDate(middle, zoneId)
+		} else if (startZonedDateTime.getDayOfYear() == 1 && stopZonedDateTime.getDayOfYear() == 1) {
+			if (startZonedDateTime.getYear() == stopZonedDateTime.getYear() - 1) {
```
mirzov, line 120: Empty argument lists on simple getters are not needed in Scala. By convention, they should only be provided if the call has side-effects.

mirzov (approved):

Issue #12, opened 2018-11-06
Station labeling app: show error messages to the user


mirzov:
Especially relevant when file uploads fail

mirzov:
Fixed in PR #93

PR #93, opened 2021-02-02
[labeling] Use toasters for giving feedback to user

roger-groth:
Simple toaster in upper right corner of window. Is used for reporting on file uploads and errors that might occur.


mirzov:
Fixes issue #12

PR #95, opened 2021-02-04
Expose metadata using schema.org

jonathanthiry:
Data object landing page include the metadata using JSON-LD  and schema.org.
A sitemap provide the list of level 2 and 3 data landing pages.


jonathanthiry:
Google requires only the name and description properties for dataset structured data. A name is always provided but some datasets are missing a description. L3 data uses the description when available. L2 data uses the specification comments. These are always provided for SITES but never for ICOS. What could we provide?

The distribution property expects a [DataDownload](https://schema.org/DataDownload) and Google recommends to provide a encodingFormat for these. An encodingFormat is a media type typically expressed using a MIME format. Is this something that we could add to our metadata?

PR #111, opened 2021-04-21
Use separate submitter config

jonathanthiry:


jonathanthiry:
The submitters.conf file would be required for the application to work. Is this good or should it defaut to an empty list?

mirzov (changes requested):
That proved to be an elegant change, nice! Good question about defaulting submitters to an empty list, I think it is a good change to add. Otherwise looks perfect to me.

mirzov (approved):

PR #119, opened 2021-06-18
[uploadgui] Add stations to production dropdowns

jonathanthiry:


mirzov:
@jonathanthiry  Please test-run this before merging. Wanted to request a review from you, but GitHub does not seem to allow that.

PR #121, opened 2021-07-27
Remove outdated Bintray resolver and use sbt default ones

jonathanthiry:


.sbtopts
```diff
@@ -1 +1 @@
--J-XX:MaxMetaspaceSize=512M -Dsbt.override.build.repos=true -Dsbt.repository.config=project/repositories
```
mirzov, line 1: This line should be kept intact. I have made some changes to Nexus configs on repo.icos-cp.eu, so it should hopefully work now. We need to rely on it in any case, because that is where our own SBT plugins are published.

project/repositories
```diff
@@ -2,4 +2,3 @@
   local
   my-ivy-proxy-releases: https://repo.icos-cp.eu/content/groups/ivy-releases, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
   my-maven-proxy-releases: https://repo.icos-cp.eu/content/groups/public
-  sbt-plugin-releases: https://dl.bintray.com/sbt/sbt-plugin-releases, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
```
mirzov, line 5: This change should be fine.

mirzov (changes requested):

.sbtopts
jonathanthiry: I think it should work either way. If we keep the override, our nexus repo acts as a proxy and would use the upstream repo if its layout policy is set up as permissive. If we remove the override, the default repos and ours would be used. I can change it back, though.

PR #136, opened 2021-11-16
Create DOI from upload gui

jonathanthiry:


src/main/scala/se/lu/nateko/cp/meta/services/upload/DoiService.scala
```diff
@@ -0,0 +1,142 @@
...
+				formats = Seq(),
+				version = Some(Version(1, 0)),
+				rightsList = Some(Seq(ccby4)),
+				descriptions = coll.description.map(d => Description(d, DescriptionType.Abstract, None)).toSeq,
+				url = Some(uri.toString)
+			)
+		}
+	}
+
+	private def fetchDataObject(uri: java.net.URI): Future[DataObject] = http
```
mirzov, line 116: We should not be making HTTP calls for this, we are already inside meta's backend. Besides, these calls will go to the production server, so will not work for resources that have been created on a dev- or staging machine.

mirzov (changes requested):

PR #158, opened 2022-06-02
Use relative urls on landing pages

jonathanthiry:


src/main/twirl/views/landpagesnips/resourceLink.scala.html
```diff
@@ -5,8 +5,8 @@
 <a href=@resUri>@{res.label.getOrElse(res.uri.toString)}</a>
 
 @resUri = @{
-	val uriStr = res.uri.toString
-	if(envri == Envri.ICOS && (uriStr.startsWith("https://meta.") || uriStr.startsWith("http://meta.")))
+	if(res.uri.getHost.startsWith("meta."))
```
mirzov, line 8: I guess we don't need implicit Envri here any more. Maybe, if it's not too much trouble, we could ask for implicit EnvriConfig here instead, and check if uri starts with dataItemPrefix or metaItemPrefix, in which case use relative link. That would prevent us from comparing with a hard-coded "meta." string, which may be wrong for some other hypothetical envri (ICOS Cities, perhaps?).

mirzov (approved):

PR #168, opened 2022-11-15
Add checkbox for automatic deprecation of files

klarakristina:


jonathanthiry:
The checkboxes were on top of each other on smaller screens, so I switched to a flex layout.

Screenshot 2022-11-16 at 07 59 28

jonathanthiry:
You could use [Bootstrap tooltips](https://getbootstrap.com/docs/5.2/components/tooltips/) for the help question marks if you think that's better in this case.

jonathanthiry (approved):

mirzov (approved):
:+1:

PR #194, opened 2023-03-28
Include both model and serial number in instrument label

klarakristina:


mirzov (changes requested):
- It would be beneficial to handle the cases when only the instrument model, or only the serial number, are available.

- It may be a good idea to use `inline def model` and `inline def serialNumber` instead of eagerly defining `val`s, as they may not be needed in case if `hasName` is available for the instrument.

mirzov (approved):

PR #222, opened 2023-09-22
Add discontinued property to SITES stations

jonathanthiry:


mirzov (changes requested):
To be consistent with the Scala code changes, the new `isDiscontinued` property in OWL should use the more specific class cpmeta:SitesStation (please check spelling) as rdfs:domain, rather than the most general cpmeta:Station. It would make sense to use cpmeta:Station if we were introducing the new property in the `Station` Scala class. Which perhaps we should do eventually, but we are not doing it in this change.

mirzov (approved):

PR #231, opened 2023-10-09
Prevent concurrent metadata upload of same object

klarakristina:


mirzov (changes requested):
`def registerEtcUpload` remained unprotected by the change. It's probably better to protect the `private def registerObjUpload` instead, as every other data/doc upload method depends on it. While at it, it could be good to protect `def registerStaticCollection`, too. It's an unlikely scenario that this will be needed, but may happen if a wrongly written client script will attempt multiple identical concurrent collection registrations. The `ObjectLock`'s message should be accordingly generalized to cover for data-, document objects, and collections.

`private[this] val` is not needed in Scala 3 any more, the compiler is smart enough to produce optimal byte code for `private val`.

mirzov (approved):

PR #265, opened 2024-05-14
Manufacturer is required for Devices

squaregoldfish:
In the metadata editor for Devices, `model` seems to be required (if that's what a yellow heading on a property indicates), but `manufacturer` is not. I've added the `manufacturer` as a limit, which I believe will make it required. I've also restricted it to Commercial Companies, as is done elsewhere in the spec.


mirzov:
To make it required, one needs to add a cardinality restriction (e.g. someValuesFrom)

PR #276, opened 2024-12-20
Remove TestDbSuites. Run Query and Route tests directly.

ggVGc:
This removes the `TestDbSuites` composite suite, which instantiates a single `TestDb` which is then shared by `SparqlRouteTests` and `SparqlTests` and cleaned up when they finish. Instead a new `TestDb` is now instantiated in each of the tests that use it.
The benefit of this is that we do not have to disable test discovery, which in my experience makes things like filtering on test name break, and makes the reporting less clear since it is always `TestDbSuites` that runs/fails and we don't see the runtime for each child test. It also seems to me that this increases parallelization, which is a benefit since these two tests seem to be the slowest in the total suite.
What we lose compared to the old version is some extra work for setting up the `TestDb` twice, however from what I see this is negligible in relation to actually running the tests.

Some rough numbers from my machine for reference:

**Old: Clean build + test**
> [info] Run completed in 23 seconds, 293 milliseconds.
[info] Total number of tests run: 281
[info] Suites: completed 37, aborted 0
[info] Tests: succeeded 281, failed 0, canceled 0, ignored 3, pending 0
[info] All tests passed.
[success] Total time: 64 s (01:04), completed Dec 20, 2024, 6:34:56 PM

**Old: Re-run test**
> [info] Run completed in 23 seconds, 266 milliseconds.
[info] Total number of tests run: 281
[info] Suites: completed 37, aborted 0
[info] Tests: succeeded 281, failed 0, canceled 0, ignored 3, pending 0
[info] All tests passed.
[success] Total time: 26 s, completed Dec 20, 2024, 6:36:57 PM

**New: Clean build + test**
> [info] Run completed in 23 seconds, 265 milliseconds.
[info] Total number of tests run: 281
[info] Suites: completed 36, aborted 0
[info] Tests: succeeded 281, failed 0, canceled 0, ignored 3, pending 0
[info] All tests passed.
[success] Total time: 45 s, completed Dec 20, 2024, 6:38:32 PM

**New: Re-run test**
> [info] Run completed in 25 seconds, 894 milliseconds.
[info] Total number of tests run: 281
[info] Suites: completed 36, aborted 0
[info] Tests: succeeded 281, failed 0, canceled 0, ignored 3, pending 0
[info] All tests passed.
[success] Total time: 28 s, completed Dec 20, 2024, 6:39:34 PM


src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/regression/QueryTests.scala
```diff
@@ -5,22 +5,17 @@ import org.eclipse.rdf4j.model.ValueFactory
...
 import org.scalatest
-import org.scalatest.BeforeAndAfterAll
-import org.scalatest.DoNotDiscover
 import org.scalatest.Informer
 import org.scalatest.compatible.Assertion
 import org.scalatest.funspec.AsyncFunSpec
 import se.lu.nateko.cp.meta.api.CloseableIterator
-import se.lu.nateko.cp.meta.test.services.sparql.TestDbFixture
-import se.lu.nateko.cp.meta.utils.rdf4j.createDateTimeLiteral
-import se.lu.nateko.cp.meta.utils.rdf4j.createLiteral
```
ggVGc, line 16: Unrelated to this change, but these imports are unused.

src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/SparqlRouteTests.scala
```diff
@@ -20,19 +19,18 @@ import se.lu.nateko.cp.meta.core.data.EnvriConfig
 import se.lu.nateko.cp.meta.core.data.EnvriConfigs
 import se.lu.nateko.cp.meta.routes.SparqlRoute
 import se.lu.nateko.cp.meta.services.sparql.Rdf4jSparqlServer
-import se.lu.nateko.cp.meta.test.services.sparql.TestDbFixture
 
 import java.net.URI
-import scala.concurrent.Await
```
ggVGc, line 26: Also unused import from before.

mirzov (approved):

PR #277, opened 2025-01-08
Refactor TestDB for readability

ggVGc:
In order to familiarize myself with the code, and to practice some Scala futures, I refactored `TestDB`. I believe this should be quite safe, as any errors should result in test breakage.
I did this mostly as an exercise, but if the changes are useful, maybe we can merge them.


jonathanthiry (commented):
Nice refactoring! I agree that it makes it a bit easier to read and understand.
I'm not sure about how helpful it is to have the `createRepo()` def instead of the code inline with the val. The comment says that the repo is created three times, but createRepo is called only once and the creations are handled inside the function. Maybe it would be useful to mark where in the for completion the three creations happen, because removing the `repo0`, `repo1`, and `repo3` vals made it less obvious.

mirzov (approved):

PR #283, opened 2025-01-28
[Suggestion/Discussion] Improve logback configuration

ggVGc:
As a followup of https://github.com/ICOS-Carbon-Portal/meta/pull/281, I think we can improve our logback configuration, finding a balance between readability and information content.
My view is that the test and production configurations serve different use-cases and can therefore be a bit different. I prefer log output during test to be terser by default, since the configuration can be manually edited in cases where needed. As a future change, I would like us to have structural logging (JSON, probably) through a secondary logback appender, but that is only relevant if we also have a way to ingest those structured logs.

My first commit here is my suggested configuration, where I have opted to remove `thread`, since I feel the Akka system is more relevant. However, it seems essentially the whole application runs in the `cpmeta` system anyway, so possibly we should change this to only having `thread` and not the actor system (since the name of the system is included in the thread name).
I also opt for `milliseconds-since-start` as the time in the test config, compared to full date in the production config. The reason being that the day, hour and minute (mostly) part of date will always be the same during a test run, and the interesting information is how much time passes between events during tests, rather than actual date. 

**Output examples:**
- Production: `meta [INFO] [2025-01-27 17:06:13,050] CitationProvider (cpmeta) - Initializing LmdbStore SailRepository...`
- Test: `[INFO] [5.482] CitationProvider (sparqlRoutingTesting) - Initializing LmdbStore SailRepository...`

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -3,7 +3,7 @@
 	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
 		<filter class="se.lu.nateko.cp.meta.HttpApiEtagWarningFilter"/>
 		<encoder>
-			<pattern>%date{ISO8601} %-5level [%thread] %X{akkaSource} - %msg%n</pattern>
+			<pattern>[%level] [%date{ISO8601}] %logger{0} \(%X{sourceActorSystem}\) - %msg %n</pattern>
 		</encoder>
 	</appender>
 
--- a/src/test/resources/logback-test.xml
+++ b/src/test/resources/logback-test.xml
@@ -2,7 +2,7 @@
 
 	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
 		<encoder>
-			<pattern>[%level] [%replace(%r){'(.*)(\d\d\d)$', '$1.$2'}] [%logger{0}] \(%X{sourceActorSystem}\) %msg %n</pattern>
+			<pattern>[%level] [%replace(%r){'(.*)(\d\d\d)$', '$1.$2'}] %logger{0} \(%X{sourceActorSystem}\) - %msg %n</pattern>
 		</encoder>
 	</appender>
 


jonathanthiry:
Not sure if I understand what you suggest here:
> My first commit here is my suggested configuration, where I have opted to remove thread, since I feel the Akka system is more relevant. However, it seems essentially the whole application runs in the cpmeta system anyway, so possibly we should change this to only having thread and not the actor system (since the name of the system is included in the thread name).

I think the other changes would be a nice improvement.

ggVGc:
@jonathanthiry I removed the `[%thread]` part from the logback configuration, in favor of `%X{sourceActorSystem}`, however I think maybe we should do the opposite on second thought.
Examples:
- `%X{sourceActorSystem}`: CitationProvider [sparqlRegrTesting] - LmdbStore initialized
- `%thread`: CitationProvider [sparqlRegrTesting-akka.actor.default-dispatcher-5] - LmdbStore initialized
- `%X{akkaSource}`: CitationProvider CitationProvider(akka://sparqlRegrTesting) - LmdbStore initialized

jonathanthiry:
Would it be possible to have a different config between production and development?
The thread information takes a lot of screen real estate but I would think that it is mostly interesting in production, no?

ggVGc:
I'm currently not aware of a way to have other configurations than "test" and "anything else", given by `logback.xml` and `logback-test.xml`. There's probably ways to configure this, but currently my understanding of sbt and the other scala tooling is not strong enough to say.
But, I'm not sure I personally think that we should have different logging format between locally running (development) and production logs. Having a different format for testing with less detailed information, however, seems reasonable.

I think, for the production logs, we err more on the side of "too much", at least until (which may be never) we have another log appended which writes JSON logs to file.

emmaicos:
Understand it as you wanted to put thread back again, no?

jonathanthiry:
I agree that the source actor system is probably not very useful. I'm actually not sure how useful the current thread information, and the dates are based on a log example using `journalctl`:
```
Sep 15 08:58:55 fsicos2 java[799033]: org.eclipse.rdf4j.rio.RDFHandlerException: java.io.IOException: Timed out trying to write data to stream
```
The date is already provided and the thread information seems to limited. Am I missing something?

PR #284, opened 2025-01-28
Fix sporadic test failures by improving TestDb initialization

ggVGc:
Sometimes our test suite fails because of what seems like initialization failure of `TestDb`, which looks like this: https://github.com/ICOS-Carbon-Portal/meta/actions/runs/12991688933/job/36229795601
This PR moves helper functions out of the `TestDb` class, leaving only initialization and cleanup, and exposes `repo` as `Repository` rather than `Future[Repository]` making it clearer where and when initialization happens.

This change results in a longer runtime of `SparqlRouteTests`, roughly from 30s to 50s on my machine, seemingly stemming from the tests relating to long running queries. However, when running tests on `master` with `~test` in sbt this error shows up after a while:
```
[ERROR] [46.751] [AbstractSail] () Closing active connection due to shut down and interrupted the owning thread of the connection Thread[#4397,se-lu-nateko-cp-meta-test-services-sparql-SparqlRouteTests-akka.actor.default-dispatcher-14,5,main] but thread is still alive after 1000 ms! 
[WARN] [46.752] [AbstractSail] () Closing active connection due to shut down; consider setting the org.eclipse.rdf4j.repository.debug system property 
[INFO] [46.758] [CoordinatedShutdown] (sparqlRoutingTesting) Running CoordinatedShutdown with reason [ActorSystemTerminateReason]
```
That behaviour is now gone (for reasons I have not dug into yet).

The rest of the test suite finishes in around 9s on my machine, both on this branch and `master`.

### DIFF
--- a/src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/regression/TestDb.scala
+++ b/src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/regression/TestDb.scala
@@ -22,6 +22,8 @@ import java.nio.file.Files
 import scala.concurrent.{ExecutionContext, Future}
 import java.nio.file.Path
 import akka.event.LoggingAdapter
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
 
 private val graphIriToFile = Seq(
 	"atmprodcsv",
@@ -52,7 +54,7 @@ class TestDb(name: String)(using system: ActorSystem) {
 	private given log: LoggingAdapter = Logging.getLogger(system, this)
 	private val dir = Files.createTempDirectory(name).toAbsolutePath
 
-	val repo: Future[Repository] =
+	val repo: Repository =
 		/**
 		The repo is created three times:
 			1) to ingest the test RDF file into a fresh new triplestore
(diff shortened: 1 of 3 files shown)


ggVGc:
Well, this test failure shows that this PR still does not fix the issue properly... https://github.com/ICOS-Carbon-Portal/meta/actions/runs/13012682459/job/36293977072?pr=284

ggVGc:
The actual issue is that `IndexHandler` doesn't support parallel usage, and deletes `sparqlMagicIndex.bin` here: https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/IndexHandler.scala#L137
So, we have a race condition when using multiple instances of `TestDb`. I still don't like the previous solution of having `QueryTests` and `SparqlRouteTests` into one suite, but currently do not have a good idea of how to share one `TestDb` instance. I tried making it into a singleton object, which I think should work, but there is some problem with the execution context, which I don't currently understand, making `TestDb.ingestTriplestore` wait forever.

The sporadic test failures are not that common, so I think we can live with it for a while until we figure out the solution.

ggVGc:
https://github.com/ICOS-Carbon-Portal/meta/pull/284/commits/09fae1c9a334e5e79acbf2c516aabb429dc8510a should fix the issue by again using a single instance of the test repository. From the perspective of the tests, things still look the same, they create an instance of `TestDb` and call `cleanup` on exit. Internally, we simply count open references, and close `TestRepo` when all references are dropped.

src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/regression/TestDb.scala
```diff
@@ -20,113 +19,142 @@ import se.lu.nateko.cp.meta.utils.async.executeSequentially
...
+private val metaConf = se.lu.nateko.cp.meta.ConfigLoader.default
+
+class TestDb {
+	TestRepo.checkout()
+
+	val repo: Repository = TestRepo.repo
+	def runSparql(query: String): Future[CloseableIterator[BindingSet]] =
+		TestRepo.runSparql(query)
+
+	override def finalize(): Unit = {
```
ggVGc, line 57: Motivation for using finalizer instead of the previous `cleanup` method:
- If tests are cancelled or forcefully restarted, the `afterAll` method from the `BeforeAndAfterAll` trait does not get triggered anyway, so cleanup does not always run. This can be verified by checking `/tmp` (at least on linux) for directories created by `TestDb`.
- During normal test runs using `test` or `~test` in sbt, the finalizer is called and cleanup happens, from my observation.
- The usage of `BeforeAndAfterAll` is what makes `SparqlRouteTest` take roughly 20s longer, even after `db.cleanup()` has been called, for reasons currently unknown to me.

PR #287, opened 2025-01-31
Add test tags: DbTest, SlowRoute

ggVGc:
This adds two test tags, allowing the exclusion of tests depending on `TestDb` and the slowest route tests.
If the `DbTest` tag is excluded, the `TestDb` will not be initialized, which decreases total test runtime by a lot. On my machine all tests finish in 5 seconds in this case.

Usage (in sbt):
- `testOnly -- -l tags.DbTest`
- `testOnly -- -l tags.SlowRoute`


ggVGc:
Merging this, as it is just a quality-of-life improvement without any functional changes other than lazy initialization of `TestDb`.

PR #290, opened 2025-02-03
Enable -Werror, -Wunused:imports and fix all warnings using scalafix

ggVGc:
This was automatically done by adding `scalafix` as in the diff below and running `scalafixAll` in `sbt`.
Source: https://scalacenter.github.io/scalafix/docs/users/installation.html

~~This PR fixes all unused imports in `.scala` files, however many of the warnings in `.scala.html` files are false, hence we cannot turn on `-Werror` yet, unfortunately.
Having `-Wunused:imports` is still useful though, since we will now get warnings from `metals` when working in `.scala` files.~~
This was fixed by: https://github.com/ICOS-Carbon-Portal/meta/pull/290/commits/8971fff5243a2e2f3dcca7d79f70310763226e91

```diff
diff --git a/build.sbt b/build.sbt
index 4bfed609..bfa21fcd 100644
--- a/build.sbt
+++ b/build.sbt
@@ -4,6 +4,8 @@ import sbt.librarymanagement.InclExclRule
 Global / onChangedBuildSource := ReloadOnSourceChanges
 ThisBuild / organization := "se.lu.nateko.cp"
 ThisBuild / scalaVersion := "3.3.4"
+ThisBuild / scalaVersion := "3.3.4"
+ThisBuild / semanticdbEnabled := true
 
 val commonScalacOptions = Seq(
 	"-encoding", "UTF-8",
diff --git a/project/plugins.sbt b/project/plugins.sbt
index add711ea..2e8ec502 100644
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,3 +8,5 @@ addSbtPlugin("se.lu.nateko.cp" % "icoscp-sbt-deploy" % "0.4.0")
 addSbtPlugin("eu.icoscp" % "icoscp-sbt-codegen" % "0.3.2")
 
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.17.0")
+
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.14.0")
```

and this `.scalafix.conf` configuration:
```
rules = [
  OrganizeImports,
  RemoveUnused
]

OrganizeImports.groupedImports = Merge
OrganizeImports.removeUnused = true
OrganizeImports.targetDialect = Scala3
```

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/routes/SparqlRoute.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/routes/SparqlRoute.scala
@@ -3,51 +3,29 @@ package se.lu.nateko.cp.meta.routes
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.http.caching.LfuCache
-import akka.http.caching.scaladsl.Cache
 import akka.http.scaladsl.marshalling.ToResponseMarshaller
-import akka.http.scaladsl.model.HttpEntity.Default
-import akka.http.scaladsl.model.HttpEntity.Strict
 import akka.http.scaladsl.model.*
+import akka.http.scaladsl.model.HttpEntity.Strict
 import akka.http.scaladsl.model.headers.*
-import akka.http.scaladsl.server.Directive
-import akka.http.scaladsl.server.Directive0
-import akka.http.scaladsl.server.Directive1
 import akka.http.scaladsl.server.Directives.*
-import akka.http.scaladsl.server.RejectionHandler
-import akka.http.scaladsl.server.RequestContext
-import akka.http.scaladsl.server.Route
-import akka.http.scaladsl.server.RouteResult
-import akka.http.scaladsl.server.RouteResult.Complete
-import akka.http.scaladsl.server.RouteResult.Rejected
+import akka.http.scaladsl.server.RouteResult.{Complete, Rejected}
 import akka.http.scaladsl.server.directives.CachingDirectives.*
-import akka.stream.Materializer
-import akka.stream.SinkShape
-import akka.stream.scaladsl.Broadcast
-import akka.stream.scaladsl.Concat
-import akka.stream.scaladsl.Flow
-import akka.stream.scaladsl.GraphDSL
-import akka.stream.scaladsl.Keep
-import akka.stream.scaladsl.Sink
-import akka.stream.scaladsl.SinkQueueWithCancel
-import akka.stream.scaladsl.Source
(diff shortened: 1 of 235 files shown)


ggVGc:
Seems some things didn't work out correctly :) Will fix that first thing tomorrow.

src/main/scala/se/lu/nateko/cp/meta/routes/SparqlRoute.scala
```diff
@@ -100,21 +78,21 @@ object SparqlRoute:
...
+						)(complete(StatusCodes.OK))
+					~
+					extractRequestContext: ctxt =>
+						//system.log.info(s"Got a SPARQL request with entity ${ctxt.request.entity}")
+						spCache.makeKey(ctxt).fold(bypass): key =>
+							cacheStatus(key): (cacheStatusMessage, cacheProhibited) =>
+								respondWithHeader(RawHeader(X_Cache_Status, cacheStatusMessage)): _ =>
+									if cacheProhibited
+									then spCache.put(key, plainRoute(ctxt))
+									else spCache.apply(key, () => plainRoute(ctxt))
```
ggVGc, line 95: This is the only change which touches non-import lines. It simply increases indentation by one step, since previously `metals` was giving an error here, and `scalafix` choked on it for the same reason. The functionality is still the same (which is covered by tests).

jonathanthiry:
Looks good! Should we add the Scalafix plugin and run it on compile as well?

ggVGc:
> Looks good! Should we add the Scalafix plugin and run it on compile as well?

I'm not sure. I think that might be better as a later step, when we've worked ourselves to enabling `-Wall`, or at least `-Wunused:all`.
There's also some decisions to make in regards to the `scalafix` configuration, and I personally don't know enough about it yet to have good ideas. I'm also not sure how "trustworthy" it is yet, so I'm a bit hesitant to add it as an automatic task.

jonathanthiry:
Yes, that's probably better to take it one step at the time. `-Wunused:imports` and `-Werror` are good starts.
The only downside is that `OrganizeImports.groupedImports = Merge` caused a lot of imports to be reorganized, and that we won't have a check in place for this.

ggVGc:
Personally I'm not too bothered about "import hygiene" in the visual sense, e.g merged or not, or sorting. Unused however is more important in my eyes, since it has the potential to highlight mistakes while developing, and that leaving unused imports can give the impression that a file has more dependencies than it actually does.

jonathanthiry (approved):

emmaicos (approved):
Nice!

PR #291, opened 2025-02-12
[uploadgui] Error message when file to upload is too large.

emmaicos:
[uploadgui] Error message when file to upload is too large.

### DIFF
--- a/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
+++ b/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
@@ -16,9 +16,7 @@ import se.lu.nateko.cp.meta.core.crypto.Sha256Sum
 object FileHasher {
 
 	val FileMaxSize = 2_000_000_000
-	val FileMaxSizeMessage = "This file could be too large to upload with this web app. " +
-		"Please consider scripted uploads as instructed here: " +
-		"https://github.com/ICOS-Carbon-Portal/meta/?tab=readme-ov-file#upload-instructions-scripting"
+	val FileMaxSizeMessage = "The file you selected is too large to upload with this form. Please use scripted uploads as described in <a class='alert-link' href='https://github.com/ICOS-Carbon-Portal/meta/?tab=readme-ov-file#upload-instructions-scripting'>our documentation</a>."
 
 	def hash(file: File): Future[Sha256Sum] = readFile(file)
 		.flatMap{buff =>
--- a/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
+++ b/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
@@ -11,8 +11,6 @@ import org.scalajs.dom.crypto
 import org.scalajs.dom.File
 import org.scalajs.dom.FileReader
 
-import scala.util.{Failure, Success}
-
 import se.lu.nateko.cp.meta.core.crypto.Sha256Sum
 
 object FileHasher {
--- a/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
+++ b/uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
@@ -17,9 +17,10 @@ import se.lu.nateko.cp.meta.core.crypto.Sha256Sum
 
 object FileHasher {
 
-	val FileMaxSize = 2000000000
-	val FileMaxSizeMessage = "This file could be too large to upload with this form. " +
-		"Please contact us if you need an alternative."
+	val FileMaxSize = 2_000_000_000
+	val FileMaxSizeMessage = "This file could be too large to upload with this web app. " +
+		"Please consider scripted uploads as instructed here: " +
+		"https://github.com/ICOS-Carbon-Portal/meta/?tab=readme-ov-file#upload-instructions-scripting"
 
 	def hash(file: File): Future[Sha256Sum] = readFile(file)
 		.flatMap{buff =>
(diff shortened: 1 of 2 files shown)


uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
```diff
@@ -40,9 +41,8 @@ class FileInput(elemId: String, cb: () => Unit){
 			_hash = fail("hashsum is being computed")
 			cb()
 		}
-		if(f.size > 2000000000) {
-			UploadApp.showAlert("This file could be too large to upload with this form. " +
-				"Please contact us if you need an alternative.", "alert alert-warning")
+		if(f.size > FileMaxSize) {
+			UploadApp.showAlert(FileMaxSizeMessage, "alert alert-warning")
```
jonathanthiry, line 45: I think we can remove this warning, we would always show the error a couple of seconds later.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/FileHasher.scala
```diff
@@ -15,13 +15,24 @@ import se.lu.nateko.cp.meta.core.crypto.Sha256Sum
 
 object FileHasher {
 
+	val FileMaxSize = 2_000_000_000
+	val FileMaxSizeMessage = "This file could be too large to upload with this web app. " +
+		"Please consider scripted uploads as instructed here: " +
+		"https://github.com/ICOS-Carbon-Portal/meta/?tab=readme-ov-file#upload-instructions-scripting"
+
```
jonathanthiry, line 22: I would probably write:
> The file you selected is too large to upload with this form. Please use scripted uploads as described in [our documentation](https://github.com/ICOS-Carbon-Portal/meta/?tab=readme-ov-file#upload-instructions-scripting).

jonathanthiry: It would be nice to use an HTML link, but you have to see if it's possible to pass it through the exception.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
jonathanthiry: Oh right, it was much faster for me. It might depend on the file size and computer. It looks like the hasher would fail for files larger than 2GB in any case, right? Should we directly show the error and not even attempt to hash?

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
```diff
@@ -36,21 +36,28 @@ class FileInput(elemId: String, cb: () => Unit){
 
 	// The event is not dispatched if the file selected is the same as before
 	fileInput.onchange = _ => file.foreach{f =>
-		if(_hash.isSuccess){
+		if(file.isSuccess){
```
ggVGc, line 39: This will always be true, since we're inside `file.foreach`. I think it'd be better to write this code like:
```
file match {
  case Success(file) => {...}
  case Failure(_) => ()
}
```
as I think the `.foreach` of `Try` is highly confusing when reading the code.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
```diff
@@ -36,21 +36,28 @@ class FileInput(elemId: String, cb: () => Unit){
...
 			UploadApp.hideAlert()
-		}
-		UploadApp.whenDone(FileHasher.hash(f)){hash =>
-			if(file.toOption.contains(f)) {
-				_hash = Success(hash) //file could have been changed while digesting for SHA-256
-				_lastModified = f.asInstanceOf[js.Dynamic].lastModified.asInstanceOf[Double]
-				cb()
+
+			UploadApp.whenDone(FileHasher.hash(f)){hash =>
+				if(file.toOption.contains(f)) {
```
ggVGc, line 56: This looks a little bit strange to me, but I'm not sure if I'm reading it correctly. Since we are inside `file.foreach`, we know that `file` is `Success(f)`, so I think `file.toOption` will always be `Some(f)`, so this line would always be true?

emmaicos: If the user is selecting a new file while we're still hashing the first file, this is false.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
emmaicos: Yes, we probably still want to check `_hash` here, but for the first selected file we want it to always be true.

emmaicos: I think we actually want this to always happen, I'll do some more testing.

uploadgui/src/main/scala/se/lu/nateko/cp/meta/upload/formcomponents/FileInput.scala
ggVGc: Ah yeah, that's true. I missed the closure.
That feels like like a case where the previous hashing should be cancelled, though? But that's a separate concern, I guess. Either way, I think it could be good to add a small comment about that in the code, since I think it'd be easy to miss again (like I just did).

emmaicos: > Either way, I think it could be good to add a small comment about that in the code, since I think it'd be easy to miss again (like I just did).

There is a comment below :)

PR #294, opened 2025-02-21
Sitemap per country

emmaicos:
Added index sitemap with links to the now separated data, collections and documents sitemaps.

Possibility to get data sitemap per country (which means countryCode of a possible station), for example with url `.../se-data-sitemap.xml`. Those are not listed in the index sitemap.

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
@@ -2,7 +2,7 @@ package se.lu.nateko.cp.meta.routes
 
 import akka.http.scaladsl.model.*
 import akka.http.scaladsl.server.Directives.*
-import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.{ExceptionHandler, Route}
 import se.lu.nateko.cp.meta.api.SparqlRunner
 import se.lu.nateko.cp.meta.core.data.{EnvriConfig, EnvriConfigs, envriConf}
 import java.net.URI
@@ -14,6 +14,14 @@ object SitemapRoute {
 	val Data = "data"
 	val Collections = "collections"
 	val Documents = "documents"
+
+	private val exceptionHandler = ExceptionHandler{
+
+		case argErr: IllegalArgumentException =>
+			complete(StatusCodes.NotFound)
+
+		case err => throw err
+	}
 
 	def apply(sparqler: SparqlRunner)(using EnvriConfigs): Route = {
 		val extractEnvri = AuthenticationRouting.extractEnvriDirective
@@ -35,7 +43,7 @@ object SitemapRoute {
 			~
 			path("""([a-z]{2})-data-sitemap.xml""".r):
 				case countryCode if CountryCode.unapply(countryCode.toUpperCase).isDefined =>
-					completeRequest(SchemaOrg.dataObjsByCountry(sparqler, countryCode.toUpperCase))
+					handleExceptions(exceptionHandler){completeRequest(SchemaOrg.dataObjsByCountry(sparqler, countryCode.toUpperCase))}
 				case _ => reject
 			~
 			path("""([a-z]{4,11})-sitemap.xml""".r):
--- a/src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
@@ -5,20 +5,52 @@ import akka.http.scaladsl.server.Directives.*
 import akka.http.scaladsl.server.Route
 import se.lu.nateko.cp.meta.api.SparqlRunner
 import se.lu.nateko.cp.meta.core.data.{EnvriConfig, EnvriConfigs, envriConf}
+import java.net.URI
 import se.lu.nateko.cp.meta.services.metaexport.SchemaOrg
+import se.lu.nateko.cp.meta.core.data.CountryCode
 
 object SitemapRoute {
+
+	val Data = "data"
+	val Collections = "collections"
+	val Documents = "documents"
+
 	def apply(sparqler: SparqlRunner)(using EnvriConfigs): Route = {
 		val extractEnvri = AuthenticationRouting.extractEnvriDirective
 
-		(get & path("sitemap.xml") & extractEnvri){ implicit envri =>
+		(get & extractEnvri){ implicit envri =>
 			given EnvriConfig = envriConf
-			complete(
-				HttpEntity(
-					ContentType.WithCharset(MediaTypes.`text/xml`, HttpCharsets.`UTF-8`),
-					views.xml.Sitemap(SchemaOrg.dataObjs(sparqler) ++ SchemaOrg.collObjs(sparqler) ++ SchemaOrg.docObjs(sparqler)).body
+			path("sitemap.xml"):
+				val sitemaps = Seq(
+					URI(s"${envriConf.metaItemPrefix}${Data}-sitemap.xml"),
+					URI(s"${envriConf.metaItemPrefix}${Collections}-sitemap.xml"),
+					URI(s"${envriConf.metaItemPrefix}${Documents}-sitemap.xml"))
+
+				complete(
+					HttpEntity(
+						ContentType.WithCharset(MediaTypes.`text/xml`, HttpCharsets.`UTF-8`),
+						views.xml.IndexSitemap(sitemaps).body
+					)
 				)
-			)
+			~
+			path("""([a-z]{2})-data-sitemap.xml""".r):
+				case countryCode if CountryCode.unapply(countryCode.toUpperCase).isDefined =>
+					completeRequest(SchemaOrg.dataObjsByCountry(sparqler, countryCode.toUpperCase))
+				case _ => reject
+			~
+			path("""([a-z]{4,11})-sitemap.xml""".r):
+				case Data => completeRequest(SchemaOrg.dataObjs(sparqler))
+				case Collections => completeRequest(SchemaOrg.collObjs(sparqler))
+				case Documents => completeRequest(SchemaOrg.docObjs(sparqler))
+				case _ => reject
 		}
 	}
+
+	private def completeRequest(schemaObjects: => Seq[URI]) (using EnvriConfig) =
+		complete(
+			HttpEntity(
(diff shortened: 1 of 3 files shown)


src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,63 @@ package se.lu.nateko.cp.meta.routes
...
 object SitemapRoute {
+
+	val Data = "data"
+	val Collections = "collections"
+	val Documents = "documents"
+
+	private val exceptionHandler = ExceptionHandler{
+
+		case argErr: IllegalArgumentException =>
+			complete(StatusCodes.NotFound)
```
emmaicos, line 21: If this would happen, the message will be `The requested resource could not be found but may be available again in the future.` instead of `The requested resource could not be found.`, which is presented if there is no path match. Not sure how we want to handle it.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,63 @@ package se.lu.nateko.cp.meta.routes
...
-		(get & path("sitemap.xml") & extractEnvri){ implicit envri =>
+		(get & extractEnvri){ implicit envri =>
 			given EnvriConfig = envriConf
-			complete(
-				HttpEntity(
-					ContentType.WithCharset(MediaTypes.`text/xml`, HttpCharsets.`UTF-8`),
-					views.xml.Sitemap(SchemaOrg.dataObjs(sparqler) ++ SchemaOrg.collObjs(sparqler) ++ SchemaOrg.docObjs(sparqler)).body
+			path("sitemap.xml"):
+				val sitemaps = Seq(
+					URI(s"${envriConf.metaItemPrefix}${Data}-sitemap.xml"),
```
emmaicos, line 33: I think we might want another prefix here, not sure about the best approach!

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,63 @@ package se.lu.nateko.cp.meta.routes
...
+
+				complete(
+					HttpEntity(
+						ContentType.WithCharset(MediaTypes.`text/xml`, HttpCharsets.`UTF-8`),
+						views.xml.IndexSitemap(sitemaps).body
+					)
 				)
-			)
+			~
+			path("""([a-z]{2})-data-sitemap.xml""".r):
```
jonathanthiry, line 44: Would it be possible to do something similar to: https://github.com/ICOS-Carbon-Portal/meta/blob/4e0528370ac19220176fc1d04ecd71ee23f0d123/src/main/scala/se/lu/nateko/cp/meta/routes/FilesRoute.scala#L11

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
jonathanthiry: It might be possible to not use this, depending on the changes suggested in the other comments.

jonathanthiry (changes requested):
It appears to work well! Hopefully we can find a better way to extract the country code, and it should be good.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
emmaicos: Yes, but then we would need to change the path from `meta.icos-cp.eu/se-data-sitemap.xml` to something like `meta.icos-cp.eu/se/data-sitemap.xml`, right? Or maybe I misunderstand.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
emmaicos: I don't think I understand how it's related. Even if I change how we get the country code from the path, I don't really like introducing `def dataObjsByCountry`, taking a String and just use it in the query without validation. But I don't need to throw an exception if the String is not valid, could return an empty Seq if that's better.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,67 @@ package se.lu.nateko.cp.meta.routes
...
+				case countryCode if CountryCode.unapply(countryCode.toUpperCase).isDefined =>
+					handleExceptions(exceptionHandler){completeRequest(SchemaOrg.dataObjsByCountry(sparqler, countryCode.toUpperCase))}
+				case _ => reject
```
ggVGc, line 48: I believe we can use pattern matching on CountryCode instead, and not deal with uppercasing here:
```suggestion
			path("""([a-z]{2})-data-sitemap.xml""".r):
				case CountryCode(code) =>
					handleExceptions(exceptionHandler) {
						completeRequest(SchemaOrg.dataObjsByCountry(sparqler, code.code))
					}
				case _ => reject
```

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
ggVGc: I think regex, prefix path, or query parameter are all valid options, but it depends on what the expected use-case is. Since it's a file, it makes sense to me to include the country-code in the name, as in the current implementation, but from a resource access perspective I would probably expect something like `icos-cp.eu/data-sitemaps/se.xml` or the other way around like you wrote @emmaicos.

However, in the current implementation at least, I think we only handle lower-case country codes, so if we stick with this, I think the regex should at least be changed to `[a-zA-Z]`.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,67 @@ package se.lu.nateko.cp.meta.routes
...
+				completeRequest(SchemaOrg.collObjs(sparqler))
+			~
+			path(DocumentsSitemap):
+				completeRequest(SchemaOrg.docObjs(sparqler))
 		}
 	}
-}
\ No newline at end of file
+
+	private def completeRequest(schemaObjects: => Seq[URI]) (using EnvriConfig) =
```
ggVGc, line 61: Should probably be named `completeSitemapRequest` to differentiate in regards to the other `complete(...)` invocation, which returns the index, but otherwise looks very similar.

src/main/scala/se/lu/nateko/cp/meta/services/metaexport/SchemaOrg.scala
```diff
@@ -64,32 +64,57 @@ object SchemaOrg:
...
 
-		val specs: Iterator[String] = sparqler.evaluateTupleQuery(SparqlQuery(specsQuery)).flatMap(b =>
-			Option(b.getValue("spec")).map(_.stringValue)
-		)
+		sparqlUriSeq(sparqler, query, "dobj")
+	end dataObjs
+
+	def dataObjsByCountry(sparqler: SparqlRunner, countryCode: String)(using envriConf: EnvriConfig): Seq[URI] =
+		if (countryCode.length != 2  || !countryCode.forall(char => char.isLetter) || !CountryCode.unapply(countryCode.toUpperCase).isDefined)
+			throw new IllegalArgumentException("Undefined country code.")
```
ggVGc, line 84: I don't think we should be throwing exceptions from Scala code, but rather return an Option. The exception handling facilities in Scala are, afaik, only in the language in order to to interop with existing java code, while in Scala it's preferred to represent errors as values.

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
```diff
@@ -2,23 +2,67 @@ package se.lu.nateko.cp.meta.routes
...
+					HttpEntity(
+						ContentType.WithCharset(MediaTypes.`text/xml`, HttpCharsets.`UTF-8`),
+						views.xml.IndexSitemap(sitemaps).body
+					)
 				)
-			)
+			~
+			path("""([a-z]{2})-data-sitemap.xml""".r):
+				case countryCode if CountryCode.unapply(countryCode.toUpperCase).isDefined =>
+					handleExceptions(exceptionHandler){completeRequest(SchemaOrg.dataObjsByCountry(sparqler, countryCode.toUpperCase))}
```
ggVGc, line 47: Related to previous comment about exceptions. We shouldn't need to handle them here, and any unexpected ones should be handled by Akka routes, I believe, which I would expect to return a 500 error to the user.

src/main/scala/se/lu/nateko/cp/meta/services/metaexport/SchemaOrg.scala
```diff
@@ -64,32 +64,57 @@ object SchemaOrg:
...
+		|order by desc(?submTime)""".stripMargin
 
-		val specs: Iterator[String] = sparqler.evaluateTupleQuery(SparqlQuery(specsQuery)).flatMap(b =>
-			Option(b.getValue("spec")).map(_.stringValue)
-		)
+		sparqlUriSeq(sparqler, query, "dobj")
+	end dataObjs
+
+	def dataObjsByCountry(sparqler: SparqlRunner, countryCode: String)(using envriConf: EnvriConfig): Seq[URI] =
+		if (countryCode.length != 2  || !countryCode.forall(char => char.isLetter) || !CountryCode.unapply(countryCode.toUpperCase).isDefined)
```
ggVGc, line 83: I think it would be better to use the pattern matching on CountryCode, as in my first suggestion, rather than calling `unapply` directly.

ggVGc (commented):
I think it would be good to add a test in a new file in `meta/test/routes/SitemapRouteTest.scala` for this route. Firstly because it's good to have tests, but also because then we have a documented place where we can see the current format of the route.

emmaicos:
> I think it would be good to add a test

Will look into this as well!

src/main/scala/se/lu/nateko/cp/meta/routes/SitemapRoute.scala
emmaicos: Removed this now and changed `dataObjsByCountry` to have `CountryCode` instead of `String` as parameter.

jonathanthiry (approved):

PR #297, opened 2025-02-25
Fix test config to run in VS Code

jonathanthiry:


ggVGc:
Is this a fix for the nullpointer exception when running tests individually? Do you know why this fixes it?

jonathanthiry:
Yes, it is a fix for the null pointer exception. I think that the working directory is not the same depending on how you run the test, but I didn't check what it is set to in this case.

PR #298, opened 2025-02-25
Reduce usage of TriplestoreConnection in favor of StatementSource

ggVGc:
This refactoring changes the companion object for `TriplestoreConnection` into one for `StatementSource`, since that is the only interface the contained methods depend on, and updates all the usage sites accordingly.
This leaves us in a situation where very few places actually require a `TripleStoreConnection` (which carries much more context), and makes the rest of the locations easier to reason about and easier to test.


src/main/scala/se/lu/nateko/cp/meta/onto/labeler/Labeler.scala
```diff
@@ -5,24 +5,26 @@ import org.eclipse.rdf4j.model.vocabulary.RDFS
...
 import se.lu.nateko.cp.meta.utils.owlapi.*
 
 import scala.jdk.CollectionConverters.IteratorHasAsScala
+import se.lu.nateko.cp.meta.instanceserver.StatementSource
 
 trait InstanceLabeler:
 
 	// rdfs:label is the default, to be overridden in some implementations
-	def getLabel(instUri: IRI)(using TSC): String =
+	def getLabel(instUri: IRI)(using TriplestoreConnection): String =
```
mirzov, line 18: Can be StatementSource

src/main/scala/se/lu/nateko/cp/meta/onto/labeler/Labeler.scala
```diff
@@ -5,24 +5,26 @@ import org.eclipse.rdf4j.model.vocabulary.RDFS
...
-	final def getRdfsLabel(instUri: IRI)(using TSC): Option[String] =
+	final def getRdfsLabel(instUri: IRI)(using StatementSource): Option[String] =
 		getValues(instUri, RDFS.LABEL).headOption.map(_.stringValue)
 
-	final def getRdfsComment(instUri: IRI)(using TSC): Option[String] =
+	final def getRdfsComment(instUri: IRI)(using StatementSource): Option[String] =
 		getValues(instUri, RDFS.COMMENT).headOption.map(_.stringValue)
 
-	final def getInfo(instUri: IRI)(using TSC) = ResourceDto(
+	final def getInfo(instUri: IRI)(using TriplestoreConnection) = ResourceDto(
```
mirzov, line 27: Can be StatementSource

mirzov (changes requested):

src/main/scala/se/lu/nateko/cp/meta/onto/labeler/Labeler.scala
ggVGc, line 27:
```suggestion
	final def getInfo(instUri: IRI)(using StatementSource) = ResourceDto(
```

PR #299, opened 2025-02-26
Valter/index associate objects keywords

ggVGc:
TODO: Description

**Unrelated changes along the way**
- Change `modForDobj` into `getDataObject`, TODO: Why
- Introduce `TripleStatement`, TODO: Describe why
- Change iteration over `queue`, TODO: why

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -100,7 +100,10 @@ final class IndexData(nObjects: Int)(
 								addStat(oe, initOk)
 								// Insert spec keywords into obj index
 								val specKeywords = StatementSource.getValues(spec, hasKeywords)
-								log.info(s"spec keywords: $specKeywords")
+								specKeywords.foreach(keyword =>
+									log.info(s"spec KW: ${keyword.stringValue()}")
+									updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
+								)
 
 							} else if (spec === oe.spec) {
 								removeStat(oe, initOk)
@@ -111,7 +114,22 @@ final class IndexData(nObjects: Int)(
 				}
 
 			case `hasAssociatedProject` =>
-				// TODO: Deal with project keywords similar to spec
+				// Associated projects should always be IRI's, so crashing on other values is okay.
+				obj match {
+					case project: IRI => {
+						val projKeywords = StatementSource.getValues(project, hasKeywords)
+						if (projKeywords.length > 0) {
+							log.info(s"project, subj: $subj")
+							getDataObject(subj).foreach { oe =>
+								projKeywords.foreach(keyword =>
+									log.info(s"KW: ${keyword.stringValue()}")
+									updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
+								)
+							}
+							log.info(s"Proj: $project, keywords: $projKeywords")
+						}
+					}
+				}
 
 			case `hasName` =>
 				getDataObject(subj).foreach { oe =>
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -81,8 +81,7 @@ final class IndexData(nObjects: Int)(
 		}
 	}
 
-	// TODO: Make private and fix IndexDataTest
-	def processTriple(statement: TripleStatement, vocab: CpmetaVocab)(using
+	private def processTriple(statement: TripleStatement, vocab: CpmetaVocab)(using
 		statements: StatementSource
 	): Unit = {
 		import statement.{subj, pred, obj, isAssertion}
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -97,27 +97,7 @@ final class IndexData(nObjects: Int)(
 								oe.spec = null
 							}
 
-							// Insert spec keywords into obj index
-							val specKeywords = StatementSource.getStringValues(spec, hasKeywords)
-							specKeywords.foreach(keyword =>
-								updateCategSet(categMap(Keyword), keyword, oe.idx, isAssertion)
-							)
-
-							val specProjects = StatementSource.getUriValues(spec, hasAssociatedProject)
-							specProjects.foreach(project => {
-								val projKeywords = StatementSource.getValues(project, hasKeywords)
-								if (projKeywords.length > 0) {
-									log.info(s"project, subj: $subj")
-									getDataObject(subj).foreach { oe =>
-										projKeywords.foreach(keyword =>
-											log.info(s"KW: ${keyword.stringValue()}")
-											updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
-										)
-									}
-									log.info(s"Proj: $project, keywords: $projKeywords")
-								}
-							})
-
+							addSpecKeywords(oe, spec, isAssertion, vocab)
 						}
 					}
 				}
@@ -126,16 +106,13 @@ final class IndexData(nObjects: Int)(
 				// Associated projects should always be IRI's, so crashing on other values is okay.
 				obj match {
 					case project: IRI => {
-						val projKeywords = StatementSource.getValues(project, hasKeywords)
+						val projKeywords = StatementSource.getStringValues(project, vocab.hasKeywords)
 						if (projKeywords.length > 0) {
-							log.info(s"project, subj: $subj")
 							getDataObject(subj).foreach { oe =>
-								projKeywords.foreach(keyword =>
-									log.info(s"KW: ${keyword.stringValue()}")
-									updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
-								)
+								projKeywords.foreach { keyword =>
+									updateCategSet(categMap(Keyword), keyword, oe.idx, isAssertion)
+								}
 							}
-							log.info(s"Proj: $project, keywords: $projKeywords")
 						}
 					}
 				}
@@ -327,6 +304,21 @@ final class IndexData(nObjects: Int)(
 
 			case _ =>
 		}
+	}
+
+	private def addSpecKeywords(oe: ObjEntry, spec: IRI, isAssertion: Boolean, vocab: CpmetaVocab)(using
+		StatementSource
+	) = {
+		val specKeywords = StatementSource.getStringValues(spec, vocab.hasKeywords)
+		val specProjects = StatementSource.getUriValues(spec, vocab.hasAssociatedProject)
+
+		val specProjKeywords = specProjects.flatMap(project => {
+			StatementSource.getStringValues(project, vocab.hasKeywords)
+		})
+
+		Set.from(specKeywords ++ specProjKeywords).foreach(keyword =>
+			updateCategSet(categMap(Keyword), keyword, oe.idx, isAssertion)
+		)
 	}
 
 	private def updateStrArrayProp(
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -98,17 +98,32 @@ final class IndexData(nObjects: Int)(
 								if (oe.spec != null) removeStat(oe, initOk)
 								oe.spec = spec
 								addStat(oe, initOk)
-								// Insert spec keywords into obj index
-								val specKeywords = StatementSource.getValues(spec, hasKeywords)
-								specKeywords.foreach(keyword =>
-									log.info(s"spec KW: ${keyword.stringValue()}")
-									updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
-								)
-
 							} else if (spec === oe.spec) {
 								removeStat(oe, initOk)
 								oe.spec = null
 							}
+
+							// Insert spec keywords into obj index
+							val specKeywords = StatementSource.getStringValues(spec, hasKeywords)
+							specKeywords.foreach(keyword =>
+								updateCategSet(categMap(Keyword), keyword, oe.idx, isAssertion)
+							)
+
+							val specProjects = StatementSource.getUriValues(spec, hasAssociatedProject)
+							specProjects.foreach(project => {
+								val projKeywords = StatementSource.getValues(project, hasKeywords)
+								if (projKeywords.length > 0) {
+									log.info(s"project, subj: $subj")
+									getDataObject(subj).foreach { oe =>
+										projKeywords.foreach(keyword =>
+											log.info(s"KW: ${keyword.stringValue()}")
+											updateCategSet(categMap(Keyword), keyword.stringValue(), oe.idx, isAssertion)
+										)
+									}
+									log.info(s"Proj: $project, keywords: $projKeywords")
+								}
+							})
+
 						}
 					}
 				}
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -61,13 +61,13 @@ final class IndexData(nObjects: Int)(
 			prop,
 			prop match {
 				/** Important to maintain type consistency between props and HierarchicalBitmaps here*/
-				case FileName        => fileNameBm
-				case FileSize        => FileSizeHierarchicalBitmap(objs)
-				case SamplingHeight  => SamplingHeightHierarchicalBitmap(objs)
-				case DataStart       => dataStartBm
-				case DataEnd         => dataEndBm
+				case FileName => fileNameBm
+				case FileSize => FileSizeHierarchicalBitmap(objs)
+				case SamplingHeight => SamplingHeightHierarchicalBitmap(objs)
+				case DataStart => dataStartBm
+				case DataEnd => dataEndBm
 				case SubmissionStart => submStartBm
-				case SubmissionEnd   => submEndBm
+				case SubmissionEnd => submEndBm
 			}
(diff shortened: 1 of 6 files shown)


src/test/resources/rdf/sparqlDbInit/netcdf.rdf
```diff
@@ -35488,7 +35488,7 @@ The STILT results are also visualized at https://stilt.icos-cp.eu/viewer/?statio
 </description>
 	<title xmlns="http://purl.org/dc/terms/">High-resolution, near-real-time fluxes over Europe from CTE-HR: anthropogenic emissions per sector 2019-11</title>
 	<hasTemporalResolution xmlns="http://meta.icos-cp.eu/ontologies/cpmeta/">hourly</hasTemporalResolution>
-	<hasKeywords xmlns="http://meta.icos-cp.eu/ontologies/cpmeta/">carbon flux</hasKeywords>
+	<hasKeywords xmlns="http://meta.icos-cp.eu/ontologies/cpmeta/">test keyword</hasKeywords>
```
ggVGc, line 35491: Not sure about his change.
I wanted to focus on a single object in the test case and clearly differentiate between object keyword and spec keyword. However, that is not how the rest of the regression route tests are written, and I wonder if the `*.rdf` fixtures loaded by `TestDb` should be representing some specific data-set?
Either way, this change doesn't break the other tests.

A better solution would be to insert the expected entries into a local test db in the test, but that would require some more refactoring of `TestDb`. So this seems like the best option currently.

src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/regression/QueryTests.scala
```diff
@@ -538,4 +540,71 @@ class QueryTests extends AsyncFunSpec {
...
+
+			// And all results should be equivalent
+			assert(findTargetObject(runSparqlSync(magicFilterQuery(specKeyword))) == Some(objResult))
+			assert(findTargetObject(runSparqlSync(magicFilterQuery(projKeyword))) == Some(objResult))
+		}
+	}
+
+	private def runSparqlSync(query: String): List[BindingSet] = {
+		Await.result(db.runSparql(query), Duration.apply(5, TimeUnit.SECONDS)).toList
+	}
```
ggVGc, line 605: I don't see a reason to run the queries asynchronously in tests, unless that's specifically what's being tests, and synchronous requests makes the test much easier to read.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -19,18 +19,26 @@ import java.time.Instant
 import scala.collection.IndexedSeq as IndSeq
 import scala.collection.mutable.{AnyRefMap, ArrayBuffer}
 
+// IndexUpdate is similar to se.lu.nateko.cp.meta.instanceserver.RdfUpdate,
+// but constrained to Rdf4jStatement rather than Statement.
+// TODO: If RdfUpdate is is changed to be more constrained, we can get rid of IndexUpdate.
+final case class IndexUpdate(
+	val statement: Rdf4jStatement,
+	val isAssertion: Boolean
```
ggVGc, line 27: In `RDF` terminology, the existence of an edge (predicate) between nodes is called an assertion. Removing it is called a rejection. Personally, I find it a bit confusing to use this terminology in the code, since assertion already has a well-established meaning.

I would prefer to look at this as "inserting or removing a statement from our data store" rather than "asserting or rejecting an RDF relationship". If others agree with this, I would like to change this to `isInsertion` or similar, as well as in `RdfUpdate`.

ggVGc:
This is done, but I wonder if we should remove the test in `QueryTests`. It is partly redundant, however we don't have another test for the magic query filtering on `hasKeyword`, so maybe it makes sense to keep it?

ggVGc:
https://github.com/ICOS-Carbon-Portal/meta/pull/299/commits/b8713df1ebe40c24eba872660da2f762b5efd539 seems to have introduces issues with the ingestion, found by not seeing the same amount of keywords in the database between fresh ingestions.
It has been reverted in https://github.com/ICOS-Carbon-Portal/meta/pull/299/commits/fa396a06d93f5e8ea88902c94b41f43a671188c8 and ingestion now seems fine again in my testing.

mirzov (changes requested):

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,15 +290,31 @@ class IndexData(nObjects: Int)(
...
+		val specKeywords = StatementSource.getStringValues(spec, vocab.hasKeywords)
+		val specProjects = StatementSource.getUriValues(spec, vocab.hasAssociatedProject)
+
+		val specProjKeywords = specProjects.flatMap(project => {
+			StatementSource.getStringValues(project, vocab.hasKeywords)
+		})
+
+		Set.from(specKeywords ++ specProjKeywords).foreach(keyword =>
+			updateCategSet(categMap(Keyword), keyword, oe.idx, isAssertion)
+		)
```
ggVGc, line 314: This could be sped up by keeping a cache of keywords per spec and project during ingestion, rather than fetching for each processed data object.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,15 +299,31 @@ class IndexData(nObjects: Int)(
...
+				getDataObject(subj).foreach { oe =>
 					updateStrArrayProp(obj, Keyword, s => Some(parseCommaSepList(s)), oe.idx, isAssertion)
 				}
 
 			case _ =>
 		}
 	}
 
-	def updateStrArrayProp(
+	private def addSpecKeywords(oe: ObjEntry, spec: IRI, isAssertion: Boolean, vocab: CpmetaVocab)(using
```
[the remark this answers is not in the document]
ggVGc: I will measure the runtime impact of the new indexing, and if it shows large degradation I will add some caching while indexing.

ggVGc:
Test of indexing with on this branch:
```
2025-03-14 17:38:13,223 INFO  [CpNotifyingSail] [cpmeta-akka.actor.default-dispatcher-5]  - Initializing Carbon Portal index...
2025-03-14 17:38:13,802 INFO  [MetaDbFactory] [cpmeta-akka.actor.default-dispatcher-7] MetaDbFactory(akka://cpmeta) - ingestion done for dcat
2025-03-14 17:38:18,809 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index received 1 million RDF assertions by now...
...
2025-03-14 17:40:17,144 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index received 47 million RDF assertions by now...
2025-03-14 17:40:17,554 WARN  [IndexData] [cpmeta-akka.actor.default-dispatcher-5]  - Value already existed: asserted value 1710064809401 of property SubmissionStart on object MO2VbtYOp2M9IFlwPbFbQEaL
2025-03-14 17:40:17,555 WARN  [IndexData] [cpmeta-akka.actor.default-dispatcher-5]  - Value already existed: asserted value 1710066012628 of property SubmissionStart on object 7avypi6iJsTydipKfCUV4p4V
2025-03-14 17:40:17,559 WARN  [IndexData] [cpmeta-akka.actor.default-dispatcher-5]  - Value already existed: asserted value 1710067206647 of property SubmissionStart on object LjTL0008TzyBWV2Xe3CuXqnt
2025-03-14 17:40:19,233 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index received 48 million RDF assertions by now...
2025-03-14 17:40:20,273 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index initialized by 48313749 RDF assertions
2025-03-14 17:40:20,278 INFO  [CpNotifyingSail] [cpmeta-akka.actor.default-dispatcher-5]  - Carbon Portal index initialized with info on 2348020 data objects

```

ggVGc:
Indexing from same `rdfStorage` in `master`

```
2025-03-14 18:07:41,352 INFO  [CpNotifyingSail] [cpmeta-akka.actor.default-dispatcher-5]  - Initializing Carbon Portal index...
2025-03-14 18:07:41,743 INFO  [MetaDbFactory] [cpmeta-akka.actor.default-dispatcher-8] MetaDbFactory(akka://cpmeta) - ingestion done for dcat
2025-03-14 18:07:42,914 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index received 1 million RDF assertions by now...
....
2025-03-14 18:08:28,465 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index received 48 million RDF assertions by now...
2025-03-14 18:08:29,088 INFO  [CpIndex] [cpmeta-akka.actor.default-dispatcher-5]  - SPARQL magic index initialized by 48313815 RDF assertions

**Comment by ggVGc** [comment]
Indexing is a fair bit slower, so adding a cache is probably worth it.
But, more importantly, the number of processed RDF assertions is not the same...

**Review by jonathanthiry — APPROVED** [review]

File: src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala [hunk 6]
```diff
@@ -272,15 +282,86 @@ final class IndexData(nObjects: Int)(
...
+	}
+
+	private def updateSpecKeywords(spec: Resource)(using vocab: CpmetaVocab)(using StatementSource): Boolean = {
+		val dataObjects: Iterator[Resource] =
+			StatementSource.getStatements(null, vocab.hasObjectSpec, spec).map(_.getSubject())
+
+		// Find out if we got any results without consuming the iterator.
+		val anyObjects = dataObjects.hasNext
+
+		dataObjects.foreach(dataObject => {
```
mirzov, line 325: This block could be written as
```scala
for dataObject <- dataObjects; oe <- getDataObject(dataObject) do
			// Casting to IRI should be fine here, since getDataObject gave us an ObjEntry
			updateDataObjectKeywords(dataObject.asInstanceOf[IRI], oe)
```
By the way, we already today have a data type (i.e. spec) that has half a million data objects: https://data.icos-cp.eu/portal/#%7B%22filterCategories%22%3A%7B%22type%22%3A%5B%22etcBioMeteoRawSeriesCsv%22%5D%7D%7D
This loop can be expensive!

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,15 +282,86 @@ final class IndexData(nObjects: Int)(
...
+	private def updateDataObjectKeywords(dataObject: IRI, oe: ObjEntry)(using vocab: CpmetaVocab)(using StatementSource) = {
+		val objectKeywords = StatementSource.getValues(dataObject, vocab.hasKeywords)
+		val (specKeywords, projectKeywords) = if (oe.spec != null) {
+			val projects = StatementSource.getUriValues(oe.spec, vocab.hasAssociatedProject)
+			(
+				StatementSource.getValues(oe.spec, vocab.hasKeywords),
+				projects.flatMap(project => { StatementSource.getValues(project, vocab.hasKeywords) })
+			)
+		} else {
+			(Set(), Set())
```
mirzov, line 344: Use `(IndexedSeq.empty, IndexedSeq.empty)` here, otherwise inferred types of `specKeywords` and `projectKeywords` get pretty crazy.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,15 +282,86 @@ final class IndexData(nObjects: Int)(
...
+		}
+
+		val keywordIndex = categMap(Keyword)
+		val newKeywords = Set.from((objectKeywords ++ specKeywords ++ projectKeywords).flatMap(parseKeywords).flatten())
```
mirzov, line 348: No argument lists after `flatten`. Also, one can call `.toSeq` on any Seq, no need to wrap in `Set.from`

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,15 +282,86 @@ final class IndexData(nObjects: Int)(
 					case _ =>
 				}
 
-			case `hasKeywords` => getDataObject(subj).foreach { oe =>
-					updateStrArrayProp(obj, Keyword, s => Some(parseCommaSepList(s)), oe.idx, isAssertion)
+			case `hasKeywords` =>
+				getDataObject(subj) match {
+					case Some(oe) => updateDataObjectKeywords(subj, oe)
```
mirzov, line 287: This handler appears to assume that the change being reacted to has already been applied to the triplestore, and can be read back from it. Is that so?

mirzov (changes requested):
Looks good overall!

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
ggVGc: The set is used for deduplication, and for quicker `newKeywords.contains` a few lines below, so I think it still makes sense?

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
ggVGc: Will fix this in further commits.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
ggVGc: Any reason for `IndexedSeq` rather than just `Seq`?

ggVGc:
I've been thinking about the correct way to go about the issue of relying on reading the inserted triple during indexing, mentioned in https://github.com/ICOS-Carbon-Portal/meta/pull/299#discussion_r2003246155, and am not sure yet, but here are some thoughts.

1. I started working on removing the expectation in https://github.com/ICOS-Carbon-Portal/meta/tree/valter/associated-keywords-indexing-no-read-committed, but I am not sure I think this is the best way. It complicates the indexing code, and since we have a `StatementSource`, it's easy that the same mistake creeps in at a later time. It would be better to let `IndexData` have the assumption that indexed triples are also available in the store.
2. The place where live updating of the index happens is here: https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/IndexHandler.scala#L66. From the [RDF4J documentation](https://docs.rdf4j.org/javadoc/4.3.10/org/eclipse/rdf4j/sail/SailConnectionListener.html), the `statementAdded` handler is called when `a statement has been added in a transaction it has registered with`, which means things should be fine if our transaction level is `READ_COMMITTED` as per https://docs.rdf4j.org/javadoc/latest/org/eclipse/rdf4j/common/transaction/IsolationLevels.html. I am not sure what transaction level we are using, though, and didn't figure out how to find it yet. Support depends on the store used, though, so we would have to look at `LMDB`.
3. Writing a test which inserts a new triple and then runs a magic query should tell us if this case is covered or not. I don't think there should be any inconsistency between runs, since `statementAdded` runs before the transaction finishes, as far as I understand. So, either we can read the inserted triple, or we cannot.
4. In the current implementation, we still have a possible inconsistent state, I believe. If the transaction gets cancelled for whatever reason, we have already updated the index. Because of this, it may be preferable to move the index update outside of the transaction anyway, which would also solve the read-committed concern.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
mirzov: Sorry, I made a typo that destroyed the meaning of what I wanted to say. There is a `.toSet` method that can be called on any `Seq`, which may be more ergonomic way that wrapping a `Seq` in `Set.from`.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
mirzov: Yes, there was a reason, actually, that is the other branch of the `if` block produces a pair of `IndexedSeq`, so using `Seq.empty` instead of `IndexedSeq.empty` widens the resulting type. But this is a very small difference, and has no practical consequence for this code, so `Seq.empty` works just as well in this case.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
mirzov: Actually, now that I gave it another look, I believe there is another way. We should consider IndexData to be another consumer of a stream of RDF updates, just like the triple store is one, sort of eventsourcing-style. In theory IndexData should not even have access to the triple store, but in practice, since it would be expensive to duplicate all the information in the index, it does, to perform some lookups in some cases. However, these lookups in the triple store should be seen as last resort, and not the first thing to reach out for.
More practically, in this case, in `updateSpecKeywords`, instead of running a triplestore lookup to list the objects belonging to the spec, one can reuse the index and get a bitmap with them very readily. Moreover, `updateSpecKeywords` is normally called as a reaction to a change in the list of keywords associated with a particular spec, and the change that we do in IndexData should be a minimal one reflecting the event, not rebuilding of the whole keyword category map once per data object belonging to the affected spec. Perhaps the way to solve this is to keep the indexing as it was, with an addition of a new map from keyword to a list of related spec, and deterring the logic of merging the direct keyword associations with the indirect ones (via specs and projects) to the filtering. This way the updates will be always cheap, and we still achieve the end goal of simplifying the SPARQL queries.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
ggVGc: Looking closer at the [documentation for isolation levels](https://docs.rdf4j.org/javadoc/latest/org/eclipse/rdf4j/common/transaction/IsolationLevels.html#NONE), it seems this should always be fine, since even with no isolation we can still read our own changes.

ggVGc:
Closed in favour of https://github.com/ICOS-Carbon-Portal/meta/pull/306

PR #300, opened 2025-02-28
validate temporal coverage when deprecating data objects

emmaicos:
validate temporal coverage when deprecating data objects


jonathanthiry:
I think it's better to call the validation in the `uploadValidator` rather than from the filename validation. However, with my change, the upload will fail if the upload doesn't have a previous version. Could you see if you could make this work? I guess that `validatePrevVers` could be a source of inspiration, but I haven't looked at it.

PR #301, opened 2025-03-05
Draft to compare with dates from file when deprecating

emmaicos:
NOTE! This is just a draft.

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/upload/validation/UploadValidator.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/upload/validation/UploadValidator.scala
@@ -20,6 +20,7 @@ import java.time.Instant
 import scala.collection.mutable.Buffer
 import scala.language.strictEquality
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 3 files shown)


src/main/scala/se/lu/nateko/cp/meta/services/upload/validation/UploadValidator.scala
```diff
@@ -45,6 +46,17 @@ class UploadValidator(servers: DataObjectInstanceServers):
 					case doc: DocObjectDto => validateDoc(doc, uploader)
 			yield dto
 
+	def validatePreviousTemporalCoverage(hash: Sha256Sum, info: UploadCompletionInfo)(using envri: Envri): Try[NotUsed] =
+		val objIri = vocab.getStaticObject(hash)
+		val validation = servers.vanillaGlobal.access:
+			for
+				given DobjConn <- metaReader.getLensForDataObj(objIri)(using envri, RdfLens.global)
+			yield
+				scoped.compareToPreviousTemporalCoverageByIri(objIri, info)
```
emmaicos, line 55: Do we always want to do this or does it depend on the type of the object for example?

PR #304, opened 2025-03-26
Enable basic RDF transaction isolation (was none)

mirzov:


src/main/scala/se/lu/nateko/cp/meta/utils/rdf4j/package.scala
```diff
@@ -61,7 +62,9 @@ extension [T](res: CloseableIteration[T])
 
 extension (repo: Repository)
 
-	def transact(action: RepositoryConnection => Unit): Try[Unit] = transact(action, None)
+	def transact(action: RepositoryConnection => Unit): Try[Unit] =
+		transact(action, Some(IsolationLevels.READ_COMMITTED))
```
ggVGc, line 66: I think we should set some isolation level (otherwise why use transactions at all), but why are we choosing `READ_COMMITTED` specifically? I would prefer to always use the strongest isolation level possible as default, unless we hit practical limitations, since resource usage is easier to fix than data corruption (even though in our case we have `rdflog`, so we are still fairly safe).
Even if we do, I think we should still use `SERIALIZABLE` as the default, and adjust each usage of `Repository.transact` to a weaker isolation level, while also making sure that the code within that transaction actually is not expecting a higher level.

I would maybe even prefer that we don't have a default isolation level at all, since that choice should depend on the logic inside the transaction, but I think defaulting to `SERIALIZABLE` is still okay.

PR #305, opened 2025-03-27
Reduce visibility of IndexData objects

ggVGc:
This removes external access to the mutable array of `ObjEntry` instances inside `IndexData`, and adds a few new methods covering the existing use-cases without allowing mutation of the array.

The main motivation is that `IndexData` relies on being the only one creating and inserting new `ObjEntry` instances, and specifically because we use the length of the array for the next id here: https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala#L298


ggVGc:
Outdated. The beetter way to do this would be to move the serialization of `IndexData` into a companion object.

PR #306, opened 2025-04-03
Filter data objects on associated keywords

ggVGc:
Remake of https://github.com/ICOS-Carbon-Portal/meta/pull/299 where we index specs per keywords and do a lookup during queries, instead of attaching all associated keywords directly to data objects during indexing.

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
@@ -20,13 +20,13 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 			else
 				val nonGeoFilts = filters.filter:
 					case gf: GeoFilter => false
-					case _             => true
+					case _ => true
 				val nonGeoBm = andFiltering(nonGeoFilts)
 				val geoBms = geoFilts.flatMap(geoFiltering(_, nonGeoBm))
 				if geoBms.isEmpty then nonGeoBm else and(geoBms ++ nonGeoBm)
 
 		case Not(filter) => this.apply(filter) match {
-				case None     => Some(emptyBitmap)
+				case None => Some(emptyBitmap)
 				case Some(bm) => Some(negate(bm))
 			}
 
@@ -34,14 +34,14 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 				case cp: ContProp => Some(data.bitmap(cp).all)
 				case cp: CategProp => cp match {
 						case optUriProp: OptUriProperty => data.categMap(optUriProp).get(None) match {
-								case None                               => None
+								case None => None
 								case Some(deprived) if deprived.isEmpty => None
-								case Some(deprived)                     => Some(negate(deprived))
+								case Some(deprived) => Some(negate(deprived))
 							}
 						case _ => None
 					}
 				case boo: BoolProperty => Some(data.boolBitmap(boo))
-				case _: GeoProp        => None
+				case _: GeoProp => None
 			}
 
 		case ContFilter(property, condition) =>
@@ -53,6 +53,15 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 				.collect { case CpVocab.DataObject(hash, _) => idLookup.get(hash) }
 				.flatten
 			Some(ImmutableRoaringBitmap.bitmapOf(objIndices*))
+
+		case CategFilter(category, values) if category == Keyword =>
+			val keywords: Seq[String] = values.collect { case kw: String => kw }
+
+			val specs: Seq[IRI] = data.getKeywordSpecs(keywords)
+			val specObjects = apply(CategFilter(Spec, specs)).getOrElse(emptyBitmap)
+
+			val perValue = data.categMap(category)
+			or(values.map(v => perValue.getOrElse(v, emptyBitmap)) :+ specObjects)
 
 		case CategFilter(category, values) =>
 			val perValue = data.categMap(category)
@@ -81,7 +90,7 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 
 	private def andFiltering(filters: Seq[Filter]): Option[ImmutableRoaringBitmap] =
 		collectUnless(filters.iterator.flatMap(this.apply))(_.isEmpty) match
-			case None      => Some(emptyBitmap)
+			case None => Some(emptyBitmap)
 			case Some(bms) => and(bms)
 
 	private def geoFiltering(
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
@@ -47,26 +47,36 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 		case ContFilter(property, condition) =>
 			Some(data.bitmap(property).filter(condition))
 
-		case CategFilter(category, values) if category == DobjUri =>
+		case CategFilter(DobjUri, values) =>
 			val objIndices: Seq[Int] = values
 				.collect { case iri: IRI => iri }
 				.collect { case CpVocab.DataObject(hash, _) => idLookup.get(hash) }
 				.flatten
 			Some(ImmutableRoaringBitmap.bitmapOf(objIndices*))
 
-		case CategFilter(category, values) if category == Keyword =>
-			val keywords: Seq[String] = values.collect { case kw: String => kw }
-			Some(data.keywordBitmap(keywords))
-
 		case CategFilter(category, values) =>
-			val perValue = data.categMap(category)
-			or(values.map(v => perValue.getOrElse(v, emptyBitmap)))
+			category match {
+				case cat : BasicCategProp =>  {
+					val perValue = data.categMap(cat)
+					or(values.map(v => perValue.getOrElse(v, emptyBitmap)))
+				}
 
-		case GeneralCategFilter(category, condition) => or(
+				case Keyword => {
+					val keywords: Seq[String] = values.collect { case kw: String => kw }
+					Some(data.keywordBitmap(keywords))
+				}
+			}
+
+		case GeneralCategFilter(category, condition) => 
+			// TODO: Enable again
+			None
+			/*
+			or(
 				data.categMap(category).collect {
 					case (cat, bm) if condition(cat) => bm
 				}.toSeq
 			)
+			*/
 
 		case gf: GeoFilter =>
 			geoFiltering(gf, None)
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
@@ -56,7 +56,7 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
 
 		case CategFilter(category, values) if category == Keyword =>
 			val keywords: Seq[String] = values.collect { case kw: String => kw }
-			Some(data.getKeywordsBitmap(keywords))
+			Some(data.keywordBitmap(keywords))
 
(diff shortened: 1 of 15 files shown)


src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
```diff
@@ -20,49 +20,71 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
...
+						}.toSeq
+					)
+				}
+
+				case Keyword => {
+					// TODO: How do I apply the filter?
+					// TODO: A test should be failing.
+					// val filtered = data.allKeywords().filter(condition)
+					val filtered: Seq[String] = Seq.empty
+					Some(data.keywordBitmap(filtered))
```
ggVGc, line 85: This needs to be implemented.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -267,19 +320,139 @@ final class IndexData(nObjects: Int)(
...
+						// TODO: If we can assume specs are always inserted before their hasKeywords triple,
+						//			 we could use the `specs` array to know if this is a spec.
+						if (isSpec) {
+							updateSpecKeywords(ensureIRI(subj), isAssertion, changedKeywords)
+						} else {
+							val project = subj
+							StatementSource.getStatements(null, vocab.hasAssociatedProject, project)
+								.map(_.getSubject())
+								.foreach(spec =>
+									// TODO: Does not cover the case where two projects have overlapping keywords
```
ggVGc, line 352: Write test for this and cover it.

ggVGc: A spec cannot have multiple projects, so this does not need to be covered.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
```diff
@@ -32,37 +32,33 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
...
-				case cp: CategProp => cp match {
-						case optUriProp: OptUriProperty => data.categMap(optUriProp).get(None) match {
-								case None                               => None
-								case Some(deprived) if deprived.isEmpty => None
-								case Some(deprived)                     => Some(negate(deprived))
-							}
-						case _ => None
-					}
+				case optUriProp: OptUriProperty => 
+					// TODO: Not covered by tests, so unsure if it's correct right now.
```
ggVGc, line 36: Add test

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -272,14 +335,122 @@ final class IndexData(nObjects: Int)(
...
+			return Set.empty;
+		}
+
+		StatementSource.getValues(subject, vocab.hasKeywords)
+			.flatMap(parseKeywords)
+			.flatten()
+			.toSet
+	}
+
+	private def setSpecKeywords(spec: IRI, keywords: Set[String]) = {
```
mirzov, line 427: This method also does not have to deal with bitmaps if we simplify the meaning of `_keywordsToSpecs` as suggested above.

mirzov (changes requested):
See the comments and an incoming PR to this branch

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -64,16 +75,69 @@ final class IndexData(nObjects: Int)(
...
+
+	def categoryBitmap(prop: CategProp, predicate: PartialFunction[prop.ValueType, Boolean]): ImmutableRoaringBitmap = {
+		categoryBitmap(prop, categoryKeys(prop).filter(predicate))
+	}
+
+	def categoryKeys(prop: CategProp): scala.collection.Set[prop.ValueType] = {
+		(prop match {
+			case Keyword =>
+				val objKeywords = categMap(Keyword).keySet
+				(_keywordsToSpecs.keySet ++ objKeywords).toSet.asInstanceOf[Set[prop.ValueType]]
```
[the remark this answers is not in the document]
ggVGc: Yeah, I agree, and I would prefer to remove all usages of it. But, I spent some time trying to understand Scala's dependent typing, and have not yet managed to figure it out without relying on typecasting.
It seems like a generally unsolved issue in this part of the code, stemming from the type class hierarchy in `DataObjectFetch.scala` and the `categMaps` map. I was trying to understand if things could be improved by using match types, but didn't get anywhere.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -64,16 +75,69 @@ final class IndexData(nObjects: Int)(
 			}
 		).asInstanceOf[HierarchicalBitmap[prop.ValueType]]
 
-	def categMap(prop: CategProp): AnyRefMap[prop.ValueType, MutableRoaringBitmap] = categMaps
-		.getOrElseUpdate(prop, new AnyRefMap[prop.ValueType, MutableRoaringBitmap])
-		.asInstanceOf[AnyRefMap[prop.ValueType, MutableRoaringBitmap]]
+	def categoryBitmap(prop: CategProp, values: Iterable[prop.ValueType]): ImmutableRoaringBitmap = {
+		prop match {
+			case Keyword =>
+				keywordBitmap(values.asInstanceOf[Seq[String]])
```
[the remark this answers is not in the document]
ggVGc: yep, sloppy error, because of the typecasting...

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -31,10 +32,14 @@ final case class StatEntry(key: StatKey, count: Int)
 def emptyBitmap = MutableRoaringBitmap.bitmapOf()
 
 final class IndexData(nObjects: Int)(
+	// These members are public only because of serialization, and should not be accessed directly.
+
 	val objs: ArrayBuffer[ObjEntry] = new ArrayBuffer(nObjects),
 	val idLookup: AnyRefMap[Sha256Sum, Int] = new AnyRefMap[Sha256Sum, Int](nObjects * 2),
+	val _specs: ArrayBuffer[IRI] = new ArrayBuffer(nObjects),
+	val _keywordsToSpecs: AnyRefMap[String, MutableRoaringBitmap] = new AnyRefMap[String, MutableRoaringBitmap](nObjects),
```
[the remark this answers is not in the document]
ggVGc: Because of memory usage and specs having multiple keywords. Basically this line: https://github.com/ICOS-Carbon-Portal/meta/pull/306/files#diff-17ac3fb53b2404c508697e4e2af36b41fc4356f18ddb75cfae479d8869bba92fR119

The lookup of objects from `categMaps` happens here: https://github.com/ICOS-Carbon-Portal/meta/pull/306/files#diff-17ac3fb53b2404c508697e4e2af36b41fc4356f18ddb75cfae479d8869bba92fR109

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -31,10 +32,14 @@ final case class StatEntry(key: StatKey, count: Int)
 def emptyBitmap = MutableRoaringBitmap.bitmapOf()
 
 final class IndexData(nObjects: Int)(
+	// These members are public only because of serialization, and should not be accessed directly.
+
 	val objs: ArrayBuffer[ObjEntry] = new ArrayBuffer(nObjects),
 	val idLookup: AnyRefMap[Sha256Sum, Int] = new AnyRefMap[Sha256Sum, Int](nObjects * 2),
+	val _specs: ArrayBuffer[IRI] = new ArrayBuffer(nObjects),
```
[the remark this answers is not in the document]
ggVGc: I'm not sure I understand how that would work, but I may be thinking incorrectly. As far as I understand, the `objs` array keeps all the objects we've seen and the `spec` member tracks which `spec` (if any) it is asssociated with. That means each spec is represented as many times as the count of the associated objects. And it doesn't give us a way to track the keywords of a spec which does not have any data objects associated with it (maybe they have not been ingested yet).
I may be seeing it from the wrong angle, but it makes sense to me to keep the list of known specs separate from the list of data objects, for the purpose of mapping keywords to specs.

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -31,8 +32,12 @@ final case class StatEntry(key: StatKey, count: Int)
 def emptyBitmap = MutableRoaringBitmap.bitmapOf()
 
 final class IndexData(nObjects: Int)(
+	// These members are public only because of serialization, and should not be accessed directly.
+
 	val objs: ArrayBuffer[ObjEntry] = new ArrayBuffer(nObjects),
 	val idLookup: AnyRefMap[Sha256Sum, Int] = new AnyRefMap[Sha256Sum, Int](nObjects * 2),
+	val specs: ArrayBuffer[IRI] = new ArrayBuffer(nObjects),
```
mirzov, line 39:
```suggestion
	val specs: ArrayBuffer[IRI] = new ArrayBuffer(100),
```

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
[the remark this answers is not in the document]
ggVGc: Changed to use simpler `AnyRefMap[String, Set[IRI]]`, since there are few enough object specs that the bitmap optimization seems superfluous.

src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/index/IndexDataTest.scala
```diff
@@ -35,4 +52,331 @@ class IndexDataTest extends AnyFunSuite {
...
+
+private def projectIRI(name: String) = {
+	factory.createIRI(s"http://meta.icos-cp.eu/resources/projects/$name")
+}
+
+private def specIRI(name: String) = {
+	factory.createIRI(s"http://meta.icos-cp.eu/resources/cpmeta/$name")
+}
+
+class StaticStatementSource(statements: Seq[Rdf4jStatement]) extends StatementSource {
```
ggVGc, line 345: Maybe use an RDF4J memory store instead?

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
```diff
@@ -32,37 +32,33 @@ class Filtering(data: IndexData, geo: Future[GeoIndex]) {
...
-			or(values.map(v => perValue.getOrElse(v, emptyBitmap)))
+			Some(data.categoryBitmap(category, values))
 
-		case GeneralCategFilter(category, condition) => or(
-				data.categMap(category).collect {
-					case (cat, bm) if condition(cat) => bm
-				}.toSeq
-			)
+		case GeneralCategFilter(category, condition) =>
+			// TODO: Not covered by tests, so unsure if it's correct right now.
```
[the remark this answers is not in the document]
ggVGc: Asana task created

src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/Filtering.scala
ggVGc: Asana task created

src/test/scala/se/lu/nateko/cp/meta/test/services/sparql/index/IndexDataTest.scala
ggVGc: Asana task created

PR #308, opened 2025-04-23
Improve robustness and indexing speed of associated keywords

mirzov:
Followup of https://github.com/ICOS-Carbon-Portal/meta/pull/306

- Changes strategy for identifying object specs, by relying on `RDF.TYPE` triples and remembering any seen specs by creating entries in `categMaps(Spec)`
- Improves tests, making sure we cover all permutations of statement processing order and that specs are not forgotten after being seen.

PR #312, opened 2025-04-25
Magic query: Unique keywords based on object filtering

ggVGc:
This PR is based on https://github.com/ICOS-Carbon-Portal/meta/pull/308, which should be merged first.

This implements a new magic query returning a list of unique keywords, given a data object filter. The only current use-case is for filtering the keywords list on https://data.icos-cp.eu/portal/

The shape of the query is: 
```
prefix cpmeta: <http://meta.icos-cp.eu/ontologies/cpmeta/>
select (cpmeta:distinct_keywords() AS ?keywords) where {
  // Data object filter
}
```

- `cpmeta:distinct_keywords()` is the special function triggering the query, and no other bindings will be set when it is used.
- The body of the query must match the shape of our magic data object query, since this implementation builds directly upon that, which also matches the use-case in the portal (filtering keywords based on filtered data objects).
- **Note:** This implementation ignores any filtering clauses that are not handled by the magic data object query. For example things like `FILTER(STRSTARTS(str(?spec), "http://meta.icos-cp.eu/"))`. This means that some keywords which do not have any results may still slip through, in some circumstances. In practice, those filtering clauses are only  used when no filters at all are applied in the portal UI, making it a fairly unimportant case. Handling such filter clauses makes the implementation a lot harder, and this level of filtering is still a huge improvement on what we had before, hence I think we can leave that as a possible future(or never) improvement.

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/CpEvaluationStrategyFactory.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/CpEvaluationStrategyFactory.scala
@@ -42,6 +42,16 @@ class CpEvaluationStrategyFactory(
 					val statsBindings = bindingsForStatsFetch(statsFetch).toIndexedSeq
 					qEvalStep(_ => statsBindings.iterator)
 
+				case UniqueKeywordsNode(bindingName, doFetch) => {
+					qEvalStep(existingBindings => {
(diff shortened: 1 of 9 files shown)


ggVGc:
This is currently deployed on `staging`.

ggVGc:
Latest commit deployed to `staging`, and seems to work based on some minimal manual testing.

PR #313, opened 2025-05-16
Benefit from explicit-nulls in CpIndex and IndexData

ggVGc:
This builds on https://github.com/ICOS-Carbon-Portal/meta/pull/310 and is the first followup-change where we remove `import scala.language.unsafeNulls` and benefit from explicit null checking.

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
@@ -15,7 +15,7 @@ import se.lu.nateko.cp.meta.services.linkeddata.UriSerializer.Hash
 import se.lu.nateko.cp.meta.services.sparql.index.*
 import se.lu.nateko.cp.meta.services.sparql.index.StringHierarchicalBitmap.StringGeo
 import se.lu.nateko.cp.meta.services.{CpVocab, CpmetaVocab}
-import se.lu.nateko.cp.meta.utils.rdf4j.{===, Rdf4jStatement, asString, toJava}
+import se.lu.nateko.cp.meta.utils.rdf4j.{Rdf4jStatement, asString, toJava}
 import se.lu.nateko.cp.meta.utils.{parseCommaSepList, parseJsonStringArray}
 
 import java.time.Instant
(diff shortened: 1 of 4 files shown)


src/main/scala/se/lu/nateko/cp/meta/services/sparql/magic/index/IndexData.scala
```diff
@@ -575,27 +574,41 @@ private def targetUri(obj: Value, isAssertion: Boolean) =
 	then obj.asInstanceOf[IRI]
 	else null
 
+// TODO: Option.scala isn't currently written for explicit-nulls. Maybe changed in later scalac versions?
+private def makeOption[A](arg: A | Null) = {
+	if (arg == null) { None }
+	else { Some(arg.nn) }
+}
```
ggVGc, line 581: This should probably be placed in some common place, since it is something that will be needed wherever we turn Java `null`s into optional values.
Alternatively, there is some standard way of doing the conversion which I am not aware of.

jonathanthiry: Wouldn't it be possible to use `Option(arg)`?

ggVGc: Nope, and the reason kind of highlights the core of what `-Yexplicit-nulls` does.

The file `Option.scala` defined this function, which is what you're referring to:
```scala
def apply[A](x: A): Option[A] = if (x == null) None else Some(x)
```
When compiled without `-Yexplicit-nulls`, this type signature is actually
```scala
def apply[A](x: A | Null): Option[A | Null]
```
because every reference type can be inhabited by `null`.

When we enable the explicit-nulls option, a type `A` actually means that type, without an implicit union with `Null`. The standard library source files are compiled with the options used for the build, so the type signature `apply(x: A)` in `Option.scala` now doesn't work for the intended use-case anymore, since the type system no longer allows us to pass in a `null` value.

In short, this is exactly the type of thing where the author of `Option.scala` should have added the correct type signature, including the union with `Null`, since the code is actually expecting `null` to be a valid input. But, instead they relied on the implicit nullability of all types, since the type system allows that (which is what I think is very broken).

ggVGc: We can of course add an overloaded extension `apply` method to `Option`, with the implementation of `makeOption`, but we still need to explicitly include that so it comes down to the same thing, just with a slightly different way of calling it.

jonathanthiry: Thanks for the explanation! This is a bit unfortunate but I now see why it is necessary.

PR #319, opened 2025-08-08
Add select none and all buttons to layer controls for station maps

ahd44:
Adds "Select none" and "Select all" buttons to the Layer Controls menu in the station map view by extending `Leaflet.Control.Layers` and adding custom CSS to the station page. Buttons are automatically hidden when 3 or fewer overlay layers are present.

Screenshot 2025-08-08 104740

---
  - https://app.asana.com/0/0/1210925718290101


jonathanthiry (commented):
It works well, nice touch with hiding the buttons when there are less than 4 layers. Did you explore [using layer groups](https://leafletjs.com/examples/layers-control/) instead?  It feels like it would have been much less custom code.

jonathanthiry (approved):
Yes, good point. Let's keep it like this, thanks!

PR #321, opened 2025-09-09
Create logback configuration for development

jonathanthiry:
Use a separate Logback configuration for development that uses colors and a simplified output. 
- Remove the project prefix name ("meta" in the example bellow)
- Remove the timestamps
- Remove the source actor
- Add color to the log level and logger

Before
Screenshot 2025-09-09 at 08 23 26

After
Screenshot 2025-09-09 at 08 17 58

This would make it easier to parse the output, and should contain the information needed most of the time.
We could also move this somewhere else to make it easier to share between projects, maybe in the infrastructure repo?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,7 +122,7 @@ lazy val meta = (project in file("."))
 			"com.typesafe.akka"     %% "akka-http-caching"                  % akkaHttpVersion cross CrossVersion.for3Use2_13,
 			"com.typesafe.akka"     %% "akka-stream"                        % akkaVersion cross CrossVersion.for3Use2_13,
 			"com.typesafe.akka"     %% "akka-slf4j"                         % akkaVersion cross CrossVersion.for3Use2_13,
-			"ch.qos.logback"         % "logback-classic"                    % "1.1.3",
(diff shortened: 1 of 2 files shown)


ggVGc:
Looks good, I would just add `[%replace(%r){'(.*)(\d\d\d)$', '$1.$2'}]` which shows [second.millisecond] since start. During development and testing I think that is more useful than full date. I think removing time information completely is bad, since it's useful being able to compare timestamps in some way between log messages.

ggVGc:
One additional comment: Removing thread and akka system information from the logs does make things a log cleaner for the general case, but `meta` actually has some concurrency going on (and some issues because of it, right now), and that is now hidden. So, that may create some confusion, but also, when working on such things, the log format can be edited locally.
Keeping it cleaner for the general case is probably preferable.

jonathanthiry:
As we talked about this in person, this configuration would only be used when running `reStart` during development, and I would argue that date/time information is not useful in this case. Good that you linked to #283, I think it is worth keeping both PRs as they are impacting separate environments, and date/time information is critical for the production and test environments.

PR #325, opened 2025-09-17
Change from imported givens to explicit calls in UriSerializer

ggVGc:
After spending some time in this code together with @jonathanlindahl this morning, I think the current implementation is a bit too confusing, and this is an attempt at a slight improvement while staying functionally the same (I believe).

### What I find confusing:
This function maps routes to different marshallings (views, basically). We can see that it takes 3 implicit (given) arguments `(Envri, EnvriConfig and ExecutionContext)`: https://github.com/ICOS-Carbon-Portal/meta/blob/d9db240b074bd9a51dffff4f88aa5b2ece7ff4c0/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala#L225
For objects and collections, it then calls https://github.com/ICOS-Carbon-Portal/meta/blob/d9db240b074bd9a51dffff4f88aa5b2ece7ff4c0/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala#L298-L300 from
https://github.com/ICOS-Carbon-Portal/meta/blob/d9db240b074bd9a51dffff4f88aa5b2ece7ff4c0/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala#L254-L259
In `delegatedRepr`, we require a given with the type `ToResponseMarshaller[() => Validated[T]]`, and in this case `T` is `StaticObject` or `StaticCollection`.

**The big question:**
Looking at `getMarshallings`, we have the 3 givens stated above, and none of them are what is expected in `delegateRepr`. Yet, we apparently have access to these givens when calling it. This means, they must be available in the outer scope of the class, or through some import mechanism, but there is no explicit such place in the file.

**The answer (partly)**
In Scala, `import` does much more than in Java, and can be used to import members from values inside scopes, rather than just types and functions from modules. Classes can have `given`'s as members and they can be public, so they can be used from outside, and hence we can also import given values. That is what happens here: https://github.com/ICOS-Carbon-Portal/meta/blob/e90d25ffd35707044588ca8e55b2ffbf8e2d10db/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala#L110-L115

**The STILL VERY CONFUSING part**
So, we now have the values `staticObjectMarshaller` and `statCollMarshaller` in the scope of the class, and the definition of the first one is https://github.com/ICOS-Carbon-Portal/meta/blob/e90d25ffd35707044588ca8e55b2ffbf8e2d10db/src/main/scala/se/lu/nateko/cp/meta/services/upload/PageContentMarshalling.scala#L29
From now on I'll focus only on staticObjectMarshaller, since the same thing applies to statCollMarshaller.
The final type of these `given`'s match what we want in `delegatedRepr`, but things don't fully make sense yet. To instantiate `staticObjectMarshaller`, we need
```scala
(using Envri, EnvriConfig, CpVocab)
```
At the point where we do
```scala
import pcm.{staticObjectMarshaller, statCollMarshaller}
```
we only have access to
```scala
given CpVocab = vocab
```
on the line above.

This means that what we have in scope (the value `staticObjectMarshaller`) is a `given`, which can be resolved if the `given`s which it depends on are also in scope, but we do not have instances of `Envri` or `EnvriConfig` in scope yet. Thise givens are actually set just before calling `getMarshallings`, here:
https://github.com/ICOS-Carbon-Portal/meta/blob/e90d25ffd35707044588ca8e55b2ffbf8e2d10db/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala#L120-L125

**Recap**
* Going back to `getMarshallings`, where we call `delegatedRepr`, we know that we have a `given` instance of `ToResponseMarshaller[() => Validated[StaticObject]]` available.
* We have `staticObjectMarshaller` in scope of the class, which is a `given` with the expected final type, but which expects 3 `givens` in order to resolve. One (`CpVocab`) is available, and two are not.
* Before calling `getMarshallings`, the other two (`Envri` and `EnvriConfig`) are set.

**The final answer**
Putting all of this together, what I believe happens is this:
* `delegatedRepr` tries to look up a `given` with the type `ToResponseMarshaller[() => Validated[StaticObject]]`
* We have such a given in scope, which is the member `staticObjectMarshaller` which is imported into scope from the value `pcm`, which is a `PageContentMarshaller`.
* When it tries to use `staticObjectMarshaller`, that in turn requires 3 more givens, and they are in scope (when called from `getMarshallings`), so the `staticObjectMarshaller` can be constructed here.

**Side note**
If you grep the code for `Validated[StaticObject]`, it may lead you to `staticObjectMarshaller` in `PageContentMarshaller`, but if you try to use go-to-references on `staticObjectMarshaller`, it will still not show you any usages. That is because the actual binding is never used, since givens are looked up by type.

### My problem with this
Fundamentally, it's very confusing to have `use ...` clauses without a clear way to see where the `given ...` statement is which fills it. If one is experienced in Scala and givens, and certain of the lookup rules, one knows that it must come from one of the imports. But, the use of `import ...` of a `given` from an object instance gives no readable indication in code of where that happens, so there is still some guess-work and jumping around to check all the imports in the scope.
In general, the combination of value imports and `given`s is a bit of a hot mess that I would prefer to avoid.

### The idea of this change
This is the most minimal change I could think of, which at least puts a clue in `UriSerializer` of where the `given`s are imported.

### Further question that I don't know the answer to
I am not sure if we even need to use a `given` here. The question I have is if, after the first time `staticObjectMarshaller` has been resolved, does it stay around in memory until next time we try to resolve it with the same `Envri`, `EnvriConfig` and `CpVocab`?
Otherwise this code can be simplified a lot. I have made a PR for that approach here: github.com/ICOS-Carbon-Portal/meta/pull/326

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala
@@ -114,10 +114,6 @@ class Rdf4jUriSerializer(
 		val stats = new StatisticsClient(config.statsClient, config.core.envriConfigs)
 		new PageContentMarshalling(config.core.handleProxies, stats)
 
-	private given CpVocab = vocab
-	import pageContentMarshalling.{given StaticObjectMarshaller}
-	import pageContentMarshalling.{given StaticCollectionMarshaller}
-
 	private val rdfMarshaller: ToResponseMarshaller[Uri] = statementIterMarshaller
 		.compose(uri => () => getStatementsIter(uri, repo))
 
@@ -253,10 +249,11 @@ class Rdf4jUriSerializer(
 
 		uri.path match
 			case Hash.Object(hash) =>
-				delegatedRepr(() => fetchStaticObj(hash))
+				given CpVocab = vocab
+				pageContentMarshalling.staticObjectMarshaller(() => fetchStaticObj(hash))
 
 			case Hash.Collection(hash) =>
-				delegatedRepr(() => fetchStaticColl(hash))
+				pageContentMarshalling.statCollMarshaller(() => fetchStaticColl(hash))
 
 			case UriPath("resources", "stations", stId) => resourceMarshallings(
 				stId, "station", fetchStation,
@@ -294,10 +291,6 @@ class Rdf4jUriSerializer(
 	end getMarshallings
 
 	private def oneOf(opts: Marshalling[HttpResponse]*): FLMHR  = Future.successful(opts.toList)
-
-	private def delegatedRepr[T](fetchDto: () => Validated[T])(
-		using trm: ToResponseMarshaller[() => Validated[T]], ctxt: ExecutionContext
-	): FLMHR = trm(fetchDto)
 
 	private def customJson[T : JsonWriter](fetchDto: () => Validated[T]): Marshalling[HttpResponse] =
 		WithFixedContentType(ContentTypes.`application/json`, () => PageContentMarshalling.getJson(fetchDto()))
--- a/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala
@@ -35,9 +35,12 @@ import se.lu.nateko.cp.meta.services.CpVocab
 import se.lu.nateko.cp.meta.services.MetadataException
 import se.lu.nateko.cp.meta.services.citation.CitationMaker
 import se.lu.nateko.cp.meta.services.citation.PlainDoiCiter
-import se.lu.nateko.cp.meta.services.upload.PageContentMarshalling
-import se.lu.nateko.cp.meta.services.upload.PageContentMarshalling.ErrorList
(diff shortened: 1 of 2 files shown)


src/main/scala/se/lu/nateko/cp/meta/services/linkeddata/UriSerializer.scala
```diff
@@ -107,12 +110,13 @@ class Rdf4jUriSerializer(
...
-	private val pcm =
+	private val pageContentMarshalling =
 		val stats = new StatisticsClient(config.statsClient, config.core.envriConfigs)
 		new PageContentMarshalling(config.core.handleProxies, stats)
 
-	given CpVocab = vocab
-	import pcm.{staticObjectMarshaller, statCollMarshaller}
+	private given CpVocab = vocab
+	import pageContentMarshalling.{given StaticObjectMarshaller}
+	import pageContentMarshalling.{given StaticCollectionMarshaller}
```
ggVGc, line 119: Maybe the type aliases make things more confusing? The important part for me is that we have `given` mentioned in the import.

jonathanthiry (approved):

PR #326, opened 2025-09-17
Alternative simplification of UriSerializer, removing `given`s

ggVGc:


jonathanthiry (approved):
Nice simplification!

ahd44:
As somone who doesn't know Scala well, this makes much more sense to me in terms of the change being made and I think it would be a lot easier to track down exactly what is being used and where. You've basically turned the implicit givens into explicit definitions, which (from my JS/TS perspective) makes it a lot easier to understand when they're imported.

jonathanthiry:
Yes, that's great. Although I forgot about your comment about what stays in memory, and I don't have an answer to that. It would be good to figure out if there are any behavior in this regard.

PR #328, opened 2025-09-24
Change download URL to use .zip file downloads from button

ahd44:
Follow-up from [data repo PR 393](https://github.com/ICOS-Carbon-Portal/data/pull/393).

The URL for the download button should be changed from `dataHost/object/HASH` to `dataHost/objects?ids=["HASH"]&fileName=filename`. The query parameters ought to be encoded in general, but our service specifically breaks if the square brackets are not encoded.

My initial change is not the best way (right way? most maintable or consistent way?) to achieve this. There should be a function added to `se.lu.nateko.cp.meta.core.data` that can be called, similarly to how `staticObjLandingPage` is called to populate `data-id` in the Add/Remove to cart buttons.

I tried to do this myself, but the URL encoding made it tricky, because the `urlEncode` defined in `meta.utils` is not available in this context; nor are the `akka` components needed to duplicate the function. I don't know quite enough about how Scala/Akka/meta work at this point to do that, but hopefully this is enough info for a backend developer to do it.

We would probably not want to alter the existing `staticObjAccessUrl` function, since it is used for the `uploadgui` to allow for uploaders to directly download the file back, but instead add a new one (e.g. `staticObjBatchAccessUrl`) that takes the `DataObject` (or `DataObject.hash` and `DataObject.fileName`) and then constructs the properly encoded URL.

Happy to work with someone else to get this done properly!

---
  - https://app.asana.com/0/0/1209841412600862

### DIFF
--- a/src/main/scala/se/lu/nateko/cp/meta/views/LandingPageHelpers.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/views/LandingPageHelpers.scala
@@ -100,6 +100,11 @@ object LandingPageHelpers:
 		s"""https://${conf.dataHost}/portal/#${params}"""
 	}
 
+	def getDownloadURL(dobj: DataObject)(using conf: EnvriConfig) = {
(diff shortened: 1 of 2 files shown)


jonathanthiry:
What do you think of my suggestion?

PR #330, opened 2025-09-30
Display all documentation links together

jonathanthiry:
The production documentation links are currently displayed in the production section, separately from the other documents. This change list them all together by using a new method that returns all documents for a given data object.

---
  - https://app.asana.com/0/0/1211438660683084

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ lazy val metaCore = (project in file("core"))
 	.enablePlugins(IcosCpSbtCodeGenPlugin)
 	.settings(
 		name := "meta-core",
-		version := "0.7.23",
+		version := "0.7.24",
 		scalacOptions ++= commonScalacOptions,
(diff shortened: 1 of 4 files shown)


jonathanthiry:
Thanks, you are suggesting some nice improvements!
- I had the same feeling about parenthesis when I saw this notation, but it looks like it is the defaut scala way now, and I don't mind it anymore. I think usually it is clear if something is a `val` or a `def`.
- I went back and forth about extracting `acquisitionDocs` into a `val`. Maybe you're right that it is easier to read like this.
- I was also used to have curly brackets, and it took me a while to addapt to the Scala 3 syntax. I would say that if an expression has an empty line it should use some kind of delimiter, and in Scala 3 it usually is `end [name]`.
In the end there are too many way to write this, I'll go for a slightly different version but incorporated most of your changes.

PR #348, opened 2026-02-25
Suggestion: Use Seq[T] instead of Option[Seq[T]] in UploadDtos

ggVGc:
Wrapping a `Seq` in `Option` has the same semantic meaning in almost every case I can think of, and in the cases where it does not, I think an alternative and more descriptive solution should probably be used.

The changes in this PR should be functionally equivalent to the previous version, while simplifying the logic using these data types, but I have not tested anything in practice yet, so there may of course be edge cases I have not thought of.
I have left comments at relevant places for why I believe these changes are functionally equivalent and valid, though.

Note: Based on https://github.com/ICOS-Carbon-Portal/meta/pull/341 since it touches same files.

- [ ] There are a few more DTOs with `Option[Seq[]]` that should probably also be changed as part of this PR.

**Testing TODO**
- Tests should be added in a new branch, based on `master`, verifying existing behavior. Then the tests can be merged into this one.
- Where could this change have visible impacts?:
    - When registering a new metadata package, do we still accept the same payloads? Parsing of a payload happens [here](https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/main/scala/se/lu/nateko/cp/meta/routes/UploadApiRoute.scala#L96).
    - The structures in [DataObject.scala)[https://github.com/ICOS-Carbon-Portal/meta/blob/master/core/src/main/scala/se/lu/nateko/cp/meta/core/data/DataObject.scala#L1] are serialized to JSON and embedded in the source of metadata previews. That is implemented in [SchemaOrg.dataJson(object: DataObject)](https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/main/scala/se/lu/nateko/cp/meta/services/metaexport/SchemaOrg.scala#L196), which can be tested directly.
    - The changes affect files in `meta/upload/subforms`. Most likely, the best way to test it is to manually check that the interface still works as expected.
- Are there existing tests, and what do they cover?
    - There does not seem to be any test covering the upload route. Adding it next to [AlternativePathRouteTest](https://github.com/ICOS-Carbon-Portal/meta/blob/master/src/test/scala/se/lu/nateko/cp/meta/test/routes/AlternativePathRouteTest.scala) would make sense.

### DIFF
--- a/core/src/main/scala/se/lu/nateko/cp/meta/core/data/DataObject.scala
+++ b/core/src/main/scala/se/lu/nateko/cp/meta/core/data/DataObject.scala
@@ -108,7 +108,7 @@ case class SpatioTemporalMeta(
 	station: Option[Station],
 	samplingHeight: Option[Float],
 	productionInfo: DataProduction,
-	variables: Option[Seq[VarMeta]]
+	variables: Seq[VarMeta]
 ){
 	def acquisition: Option[DataAcquisition] = station.map{
 		DataAcquisition(_, None, Some(temporal.interval), None, None, samplingHeight)
@@ -166,7 +166,7 @@ case class DataObject(
 
 		val varsAndPosits = for
 			cols <- specificInfo.fold(
-				l3 => l3.variables,
+				l3 => Some(l3.variables),
 				l2 => l2.columns
 			).toSeq
 			col <- cols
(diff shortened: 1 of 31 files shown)


ggVGc (commented):
Reasonings behind the main changes.

jonathanthiry (commented):
In general, I agree with the idea. I don't know for sure why it was decided to use `Option[Seq[T]]` to start with. As you are hinting in your comments, these changes would require proper testing of the impacted code paths to avoid introducing any bugs, as it is not easy to judge this at a glance.

Jonathan-Schenk (commented):
The changes you suggest make sense and improve the clarity of the code. I think it is a good idea to apply them after testing that nothing breaks because of them, although it seems that all current behaviors would be maintained.

PR #352, opened 2026-03-27
Add Sentry SDK

jonathanthiry:
This change adds the Sentry SDK and initializes it with the default configuration and uncaught route exceptions.

The Sentry DSN can be specified by adding the following line to `application.conf`.
```
cpmeta.sentry.dsn = "https://<key>@<sentry-host>/<project-id>"
```


ggVGc (approved):
Looks like it should do what we want. If there are issues, they should show up quickly during startup.

ggVGc (approved):

PR #356, opened 2026-04-02
Simplify source structure

ggVGc:
Scala does not require a match between directory structure and package name, and we currently have many nested empty directories for this reason.
These changes do not break or change the build in any way, but simplifies the file structure of the repository.


jonathanthiry (approved):
Nice that you snuck in removing the symlinks, good call ;)
Any reason why you didn’t change the structure for the two java files? It looks like it would work as well, no?

ggVGc:
> Nice that you snuck in removing the symlinks, good call ;) Any reason why you didn’t change the structure for the two java files? It looks like it would work as well, no?

Yeah, I think the symlink situation is highly related, and stumbled upon it while doing these changes, and it simplifies things. If we encounted IDE-related issues again (which seemingly was the original issue) we can find another approach.
 
Regarding java, I think I was under the impression that java sources still needed the directory structure to match the package name, but that seems not to be the case.

PR #357, opened 2026-04-22
Add a SPARQL link for types in the editor

jonathanthiry:
This adds a button in the metadata editor that link to a SPARQL query that should give a list of individuals for the currently selected type. The button can be seen at https://metastaging.icos-cp.eu/edit/cpmeta/ right now and looks like:
SPARQL button

The important parts to review would be the main logic `src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala` and the tests `src/test/scala/se/lu/nateko/cp/meta/test/InstOntoTests.scala` and `src/test/scala/se/lu/nateko/cp/meta/test/routes/MetadataEntryRoutingTests.scala`.

### DIFF
--- a/src/main/js/metaentry/stores/IndividualsStoreFactory.js
+++ b/src/main/js/metaentry/stores/IndividualsStoreFactory.js
@@ -1,4 +1,4 @@
-module.exports = function(Backend, chooseTypeAction, chooseIndividAction, createIndividualAction, deleteIndividualAction){
+module.exports = function(Backend, selectTypeAction, selectIndividAction, createIndividualAction, deleteIndividualAction){
 	return Reflux.createStore({
 
 		publishState: function(){
@@ -12,49 +12,66 @@ module.exports = function(Backend, chooseTypeAction, chooseIndividAction, create
 		init: function(){
 			this.state = {
 				individuals: [],
+				individualsSparql: null,
+				selectedType: null,
 				addingInstance: false, //needed by the IndividualsList view to hide the IndividualAdder when refreshing
-				chosen: null
+				selectedIndividual: null
 			};
-			this.listenTo(chooseTypeAction, this.fetchIndividuals);
-			this.listenTo(chooseIndividAction, this.updateChosenIndivid);
+			this.listenTo(selectTypeAction, this.fetchIndividuals);
+			this.listenTo(selectIndividAction, this.updateSelectedIndivid);
 			this.listenTo(createIndividualAction, this.createIndividual);
 			this.listenTo(deleteIndividualAction, this.deleteIndividual);
 		},
 
-		fetchIndividuals: function(chosenType){
+		fetchIndividuals: function(selectedType){
 
 			var self = this;
-			self.chosenType = chosenType;
+			self.selectedType = selectedType;
+			self.state.selectedType = selectedType;
+			self.publishState();
 
-			Backend.listIndividuals(chosenType).then(
+			Backend.listIndividuals(selectedType).then(
 				function(individuals){
 
-					if(chosenType !== self.chosenType) return;
+					if(selectedType !== self.selectedType) return;
 
 					self.state.individuals = individuals;
 					self.publishState();
 				},
 				function(err){
-					self.chosenType = undefined;
+					self.selectedType = undefined;
+					console.log(err);
+				}
+			);
+
+			Backend.getIndividualsSparql(selectedType).then(
+				function(individualsSparql){
+
+					if(selectedType !== self.selectedType) return;
+
+					self.state.individualsSparql = individualsSparql;
+					self.publishState();
+				},
+				function(err){
 					console.log(err);
 				}
 			);
 		},
 
-		updateChosenIndivid: function(chosenIndivid){
-			this.state.chosen = chosenIndivid;
+		updateSelectedIndivid: function(selectedIndivid){
+			this.state.selectedIndividual = selectedIndivid;
 			this.publishState();
 		},
 
 		createIndividual: function(newIndReq){
 			var self = this;
-			if(this.chosenType !== newIndReq.type) return;
+			if(this.selectedType !== newIndReq.type) return;
 
 			Backend.createIndividual(newIndReq.uri, newIndReq.type).then(
 				function(){
 					//TODO Revise the next two lines and the related data flow
 					self.fetchIndividuals(newIndReq.type);
-					chooseIndividAction(newIndReq.uri); //will trigger both this store and the EditStore
+					selectIndividAction(newIndReq.uri); //will trigger both this store and the EditStore
 				},
 				function(err){console.log(err);}
 			);
@@ -68,9 +85,9 @@ module.exports = function(Backend, chooseTypeAction, chooseIndividAction, create
 						return ind.uri !== indUri;
 					});
 					//TODO Revise the next lines and the related data flow
-					if(self.state.chosen === indUri) {
-						self.state.chosen == null;
-						chooseIndividAction(null);
+					if(self.state.selectedIndividual === indUri) {
+						self.state.selectedIndividual == null;
+						selectIndividAction(null);
 					} else self.publishState();
 					
 				},
@@ -80,4 +97,3 @@ module.exports = function(Backend, chooseTypeAction, chooseIndividAction, create
 
 	});
 }
-
--- a/src/main/js/metaentry/stores/IndividualsStoreFactory.js
+++ b/src/main/js/metaentry/stores/IndividualsStoreFactory.js
@@ -33,6 +33,7 @@ module.exports = function(Backend, selectTypeAction, selectIndividAction, create
 			Backend.listIndividuals(selectedType).then(
 				function(individuals){
 
+					// Ignore stale async responses after the user switched type.
 					if(selectedType !== self.selectedType) return;
 
 					self.state.individuals = individuals;
@@ -47,6 +48,7 @@ module.exports = function(Backend, selectTypeAction, selectIndividAction, create
 			Backend.getIndividualsSparql(selectedType).then(
 				function(individualsSparql){
 
+					// Ignore stale async responses after the user switched type.
 					if(selectedType !== self.selectedType) return;
 
 					self.state.individualsSparql = individualsSparql;
--- a/src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
+++ b/src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
 		val labeler = onto.getLabelerForClassIndividuals(classUri)
 		getPropValueHolders(RDF.TYPE, classUri.toRdf).map(labeler.getInfo)
 
+	def getIndividualsSparql(classUri: URI, subjectPrefix: Option[String]): String =
+		val classInfo = onto.getClassInfo(classUri)
+		val propUris = (
+			classInfo.properties.map {
+				case p: DataPropertyDto => p.resource.uri
+				case p: ObjectPropertyDto => p.resource.uri
+			} ++ Seq(RDFS.LABEL.toJava, RDFS.COMMENT.toJava)
+		).distinct.filterNot(_ == RDF.TYPE.toJava)
+
+		val usedNames = scala.collection.mutable.Set.empty[String]
+		def uniqueVarName(propUri: URI): String =
+			val base = sparqlVarBase(propUri)
+			val name = Iterator.from(1)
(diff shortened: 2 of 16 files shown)


src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+			usedNames += name
+			name
+
+		val bindings = propUris.map(uri => (uri, uniqueVarName(uri)))
+		val selectVars = ("?s" +: bindings.map{case (_, name) => s"?$name"}).mkString(" ")
+		val optionalPatterns = bindings
+			.map{case (uri, name) => s"\tOPTIONAL { ?s ${sparqlPredicate(uri)} ?$name . }"}
+			.mkString("\n")
+		val subjectPrefixFilter = subjectPrefix
+			.map(prefix => s"\tFILTER(STRSTARTS(STR(?s), ${sparqlStringLiteral(prefix)}))")
```
Jonathan-Schenk: Why not defining `subjectPrefix` as `Option[URI]` instead of `Option[String]`? It would prevent having to handle special characters like what is done here for backslashes and double-quotations. The only usage of `getIndividualsSparql` [in MetadataEntryRouting](https://github.com/ICOS-Carbon-Portal/meta/pull/357/changes#diff-2b49189007f3a723b1743da9be6d24599f282deb57c8f5642723a1d7ba05012dR49) takes as parameter an `URI` that was converted to `String`, so it is already a well-formed URI.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+			.getOrElse("")
+
+		s"""PREFIX cpmeta: <http://meta.icos-cp.eu/ontologies/cpmeta/>
+		|PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+		|PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
```
Jonathan-Schenk: `RDF.NAMESPACE` and `RDFS.NAMESPACE` could be used instead of the full namespace URI in each case. For the `cpmeta` prefix, `OntoConstants.CpmetaPrefix` can be used. This also applies to [lines 88-89](https://github.com/ICOS-Carbon-Portal/meta/pull/357/changes#diff-c71a59be447288a9541b7fed7e9a120a7c05ffb669317e45671ca7e2f4793952R88-R89).

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
 		val labeler = onto.getLabelerForClassIndividuals(classUri)
 		getPropValueHolders(RDF.TYPE, classUri.toRdf).map(labeler.getInfo)
 
+	def getIndividualsSparql(classUri: URI, subjectPrefix: Option[String]): String =
+		val classInfo = onto.getClassInfo(classUri)
+		val propUris = (
+			classInfo.properties.map {
+				case p: DataPropertyDto => p.resource.uri
+				case p: ObjectPropertyDto => p.resource.uri
+			} ++ Seq(RDFS.LABEL.toJava, RDFS.COMMENT.toJava)
```
Jonathan-Schenk: Should `RDFS.SEEALSO` also be included? In addition, there might be more annotations/properties that would need to be added for completeness and are not findable with `onto.getClassInfo`, although I have not checked it thoroughly and it might change over time.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+			else s"<$str>"
+		else if str.startsWith(rdfsPrefix) then
+			val suffix = str.substring(rdfsPrefix.length)
+			if suffix.matches("[A-Za-z_][A-Za-z0-9_\\.-]*") then s"rdfs:$suffix"
+			else s"<$str>"
+		else s"<$str>"
+
+	private def sparqlVarBase(propUri: URI): String =
+		if propUri == RDFS.LABEL.toJava then "label"
+		else if propUri == RDFS.COMMENT.toJava then "comment"
```
Jonathan-Schenk: These two cases are covered by the general one below, so they could be removed.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+		val str = uri.toString
+		if str.startsWith(cpmetaPrefix) then
+			val suffix = str.substring(cpmetaPrefix.length)
+			if suffix.matches("[A-Za-z_][A-Za-z0-9_\\.-]*") then s"cpmeta:$suffix"
+			else s"<$str>"
+		else if str.startsWith(rdfsPrefix) then
+			val suffix = str.substring(rdfsPrefix.length)
+			if suffix.matches("[A-Za-z_][A-Za-z0-9_\\.-]*") then s"rdfs:$suffix"
+			else s"<$str>"
+		else s"<$str>"
```
Jonathan-Schenk: This could be generalized with something like that.
```suggestion
		val str = uri.toString
		val prefixes: Map[String, String] = Map(
			"cpmeta" -> OntoConstants.CpmetaPrefix,
			"rdfs" -> RDFS.NAMESPACE)
		prefixes.collectFirst{
			case (prefix, namespace) if str.startsWith(namespace) =>
				val suffix = str.substring(namespace.length)
				if suffix.matches("[A-Za-z_][A-Za-z0-9_\\.-]*") then s"$prefix:$suffix"
				else s"<$str>"
		}.getOrElse(s"<$str>")
```

Jonathan-Schenk (commented):
Nice new functionality! I have only reviewed the parts that you mentioned and have a few minor questions or suggestions.

src/main/js/metaentry/stores/IndividualsStoreFactory.js
```diff
@@ -12,49 +12,66 @@ module.exports = function(Backend, chooseTypeAction, chooseIndividAction, create
...
-					self.chosenType = undefined;
+					self.selectedType = undefined;
+					console.log(err);
+				}
+			);
+
+			Backend.getIndividualsSparql(selectedType).then(
+				function(individualsSparql){
+
+					if(selectedType !== self.selectedType) return;
```
ggVGc, line 52: What case is this check handling?

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+	def getIndividualsSparql(classUri: URI, subjectPrefix: Option[String]): String =
+		val classInfo = onto.getClassInfo(classUri)
+		val propUris = (
+			classInfo.properties.map {
+				case p: DataPropertyDto => p.resource.uri
+				case p: ObjectPropertyDto => p.resource.uri
+			} ++ Seq(RDFS.LABEL.toJava, RDFS.COMMENT.toJava)
+		).distinct.filterNot(_ == RDF.TYPE.toJava)
+
+		val usedNames = scala.collection.mutable.Set.empty[String]
```
ggVGc: Do we need the full qualification here?

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+		|PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+		|PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+		|
+		|SELECT $selectVars
+		|WHERE {
+		|	?s rdf:type <${classUri.toString}> .
+		|$subjectPrefixFilter
+		|$optionalPatterns
+		|}
+		|ORDER BY ?s""".stripMargin
```
ggVGc: I think we could use [SparqlBuilder](https://rdf4j.org/documentation/tutorials/sparqlbuilder/) for SPARQL generation? I tried by adding `rdf4j-sparqlbuilder` to `build.sbt` and seems to work fine. Hopefully it could also remove the need of dealing with escaping, and we could also start using it elsewhere where we generate SPARQL queries if suitable.

src/test/scala/se/lu/nateko/cp/meta/test/InstOntoTests.scala
```diff
@@ -31,4 +33,45 @@ class InstOntoTests extends AnyFunSpec{
...
+
+			assert(query.contains("path\\\\with\\\\slash"))
+			assert(query.contains("\\\"quoted\\\""))
+		}
+
+		it("produces a syntactically valid SPARQL query"){
+			val query = instOnto.getIndividualsSparql(stationClassUri, Some("http://meta.icos-cp.eu/"))
+
+			(new SPARQLParser).parseQuery(query, null)
+			assert(true)
```
ggVGc: We should at least check that the result is a `ParsedQuery`, rather than relying on "not crash" means successful parse.
I would even prefer that we check the content of the parsed query matches what we expect.

src/test/scala/se/lu/nateko/cp/meta/test/InstOntoTests.scala
```diff
@@ -31,4 +33,45 @@ class InstOntoTests extends AnyFunSpec{
...
+	describe("getIndividualsSparql"){
+
+		it("includes basic query structure and common projected properties"){
+			val query = instOnto.getIndividualsSparql(stationClassUri, None)
+
+			assert(query.contains("SELECT ?s"))
+			assert(query.contains(s"?s rdf:type <$stationClassUri> ."))
+			assert(query.contains("ORDER BY ?s"))
+			assert(query.contains("rdfs:label"))
+			assert(query.contains("rdfs:comment"))
```
ggVGc: We could parse the query here, and assert on the content, rather than doing string interrogation. That would remove the need for the `produces a syntactically valid SPARQL query` test, since we are doing the same thing here.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
```diff
@@ -44,6 +44,85 @@ class InstOnto (instServer: InstanceServer, val onto: Onto):
...
+
+		val usedNames = scala.collection.mutable.Set.empty[String]
+		def uniqueVarName(propUri: URI): String =
+			val base = sparqlVarBase(propUri)
+			val name = Iterator.from(1)
+				.map(i => if i == 1 then base else s"$base$i")
+				.find(name => !usedNames.contains(name))
+				.get
+			usedNames += name
+			name
```
ggVGc: I think we can replace this with a `VarNameGenerator` (for lack of a better term), which holds a map with counters for each base term and outputs the next term each time it is called.

ggVGc (commented):
Some comments, but generally the functionality looks good.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
ggVGc: Yeah, I agree with this. I don't see a reason currently for it being a `String`. Technically, prefixes are `IRI`s, which is a superset of `URI`, so there are theoretical issues. But, I think we're converting between the two in many places anyway, so that's a broader concern (which doesn't seem to matter currently).

src/main/js/metaentry/stores/IndividualsStoreFactory.js
jonathanthiry: It ignores the response if the type was changed between the call and the response came.

src/main/scala/se/lu/nateko/cp/meta/onto/InstOnto.scala
jonathanthiry: I won't add this in this PR, but we could have a separate discussion about it.

ggVGc (approved):
Looks fine, but as you mentioned in the comment, I think we should look into followup work of moving to https://rdf4j.org/documentation/tutorials/sparqlbuilder/ for generating SPARQL within `meta`

Jonathan-Schenk (approved):

PR #359, opened 2026-05-04
[Unchained] Remove the magic index

ggVGc:
This simply removes the magic index and all related code. It is the starting point from which we can rework `meta` to work with an external triplestore.
We could do it the other way around, but I believe this way is the simplest.


src/main/bash/patchIndexDump.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
-# patches URL for Trebon station to be based on proper id CZ-wet instead of CZ-Wet
-cp sparqlMagicIndex.bin sparqlMagicIndex.bin.bak
-cat sparqlMagicIndex.bin.bak | gunzip > sparqlMagicIndex.bin.unz
-sed -i 's/ES_CZ-Wet/ES_CZ_wet/g' sparqlMagicIndex.bin.unz
-cat sparqlMagicIndex.bin.unz | gzip > sparqlMagicIndex.bin
```
ggVGc, line 6: Do we maybe need to do this patching on the actual triplestore?

PR #360, opened 2026-05-04
Update station pages for new website design

ahd44:
Update the station pages for ICOS stations to match the new website design, while maintaining SITES stations' layouts (with some minor tweaks to improve consistency with SITES website theme).


jonathanthiry (commented):
Good layout improvements overall!
- It's a bit unfortunate to introduce separate layouts per ENVRI. Could we look how it would work to use the same layout for SITES in a follow-up task?
- A couple of things that I know you got from the design but that I think could be improved:
  - It looks a bit weird to have the name of the station used both for the `h1` and the first `h2`.
  - I don't think the `learn more` make sense for many of the link boxes. We would need to have custom buttons or simply remove them.

ahd44:
> Good layout improvements overall!
> 
> * It's a bit unfortunate to introduce separate layouts per ENVRI. Could we look how it would work to use the same layout for SITES in a follow-up task?

Yes, definitely. There were some breaking differences on the SITES side that caused me to want to keep its layout separate for now -- mostly the image header being centered and narrower on the SITES pages.

I would be willing to continue to work on this for a bit longer if you think it's important to have done before merging.
 
> * A couple of things that I know you got from the design but that I think could be improved:   
>    * It looks a bit weird to have the name of the station used both for the `h1` and the first `h2`.
>    * I don't think the `learn more` make sense for many of the link boxes. We would need to have custom buttons or simply remove them.

Yeah, I understand. I've removed the `h2` as I don't think it's adding anything. I agree re: the "Learn more" buttons; I've removed them and replaced them with a right-arrow icon, which I think communicates it's a link clearly without needing a button (or needing specific text for each link box, which works fine on the Drupal site but not so well here).

image

jonathanthiry (approved):
Thanks, that's great!

PR #361, opened 2026-05-04
`meta-virtuoso` - External triplestore and related work

ggVGc:
This is the collection branch for the work of moving `meta` away from custom indexing and towards using an external triplestore, which initially will be `Virtuoso`.
All specific work will happen in other PRs which will be merged here, until we have a service that replicates current behaviour.


ggVGc:
@jonathanthiry This is still under active development, but at this stage it has the main changes of removing the magic index and transitioning to a new triplestore, and I think it would be good to take a quick overview look at least to see if I have made any major mistake.
The proper review should happen later on when everything is in place, but still good to get some other eyes on it a few times before then. Either way, the final testing, comparing to current production, should surface if there are any major inconsistencies at that point.

PR #364, opened 2026-05-06
[Unchained] External triplestore

ggVGc:
Mostly works, but still improvements to be made.
TODO: Summarize changes and motivation.

**Contains two scripts in `tools/` for populating Virtuoso from RDFLog:**
- One which exports `.ttl` files which could be bulk loaded directly by Virtoso.  
- Another which uses [Graph Store protocol](https://www.w3.org/2009/sparql/docs/http-rdf-update/) to send triples over HTTP.

Both of these require loading all triples from each graph in RDFLog into memory first, in order to resolve all assertions and retractions before outputting the final set of triples. This requires a lot of ram, around 10GB for me, but is still within reason. Using the normal RDF Update protocol with Virtuoso works, but only for batches smaller than a few hundred triples at a time, which is very slow for a full re-ingestion. It is also wasteful since a lot of the assertions are later retracted.

**Issues:**
- [x] ~~Currently initial fresh ingestion from RDFLog into Virtuoso fails with permission issues like: `Virtuoso 42000 Error SR186:SECURITY: No permission to execute procedure DB.DBA.SPARQL_INSERT_DICT_CONTENT` even when using `dba` user.~~
  - This was solved by https://github.com/ICOS-Carbon-Portal/meta/pull/364/changes#diff-e1f2b5f8f0029f60bb270bc106c7191c75eee54f8680f68d7c8324985d89e9cdR28-R31


src/main/scala/MetaDb.scala
```diff
@@ -77,8 +77,6 @@ object MetaDb:
 				val writeCtxt = getInstServerContext(dataObjServers, servDef)
 				servDef.label -> InstanceServerConfig(
 					logName = Some(servDef.label),
-					skipLogIngestionAtStart = servDef.replayLogFrom.map(_ => false),
-					logIngestionFromId = servDef.replayLogFrom,
```
ggVGc, line 81: Seems we should take this into account when populating from RDFLog.

src/main/resources/application.conf
```diff
@@ -375,6 +374,7 @@ cpmeta{
 		style = "elsevier-harvard"
 		eagerWarmUp = true
 		timeoutSec = 3
+		derivedCitationsGraph = "http://meta.icos-cp.eu/derived/citations/"
```
ggVGc: This will be relevant in followup PR for populating citations.

src/main/scala/persistence/postgres/PostgresRdfLog.scala
```diff
@@ -153,4 +153,8 @@ object PostgresRdfLog{
 			factory = factory
 		)
 
+	def checkConnection(conf: RdflogConfig): Unit =
+		val conn = Postgres.getConnection(conf.server, conf.credentials).get
+		conn.close()
```
ggVGc, line 158: I don't particularly like this, but it's the quickest way to make sure we have RDFLog connectivity early on during startup.

src/main/scala/MetaDb.scala
ggVGc: On closer look, this is no longer relevant. It was added in https://github.com/ICOS-Carbon-Portal/meta/commit/5cd4ceedf5ee426aefa37f9b1e027ff2f9717640, in order to continue a partial ingestion from a specific graph. If we would want this functionality again, we would add it to the `populateTriplestore` script in `tools/`.

src/main/scala/MetaDb.scala
```diff
@@ -55,12 +51,16 @@ class MetaDb (
 	val uriSerializer: UriSerializer =
 		new Rdf4jUriSerializer(vanillaRepo, vocab, metaVocab, lenses, citer.doiCiter, config)
 
-	def makeReadonlyDumpIndexAndCaches(msg: String): Future[String] =
-		repo.getSail match
-			case cp: EnrichingSail =>
-				val exe = summon[ActorSystem].dispatcher
-				cp.makeReadonlyAndDumpCaches(msg)(using exe)
-			case _ => Future.successful("Not a Carbon Portal repository, cannot switch to read-only mode")
+	def dumpIndexAndCaches(msg: String): Future[String] =
```
ggVGc, line 54: This logic was originally from https://github.com/ICOS-Carbon-Portal/meta/blob/396e0be1c3c3272dbfca90c6a013d56434cece0c/src/main/scala/services/sparql/enrichment/EnrichingSail.scala#L49-L63

PR #366, opened 2026-05-19
Add operationalPeriod to OTC station flow

jonathanthiry:
OTC will be able to provide an operational period string in https://meta.icos-cp.eu/edit/otcentry/ which will end up in the production metadata https://meta.icos-cp.eu/edit/icosmeta/. The property will be visible on the station pages and available in SPARQL using the `http://meta.icos-cp.eu/ontologies/cpmeta/hasOperationalPeriod` predicate.

### DIFF
--- a/src/main/twirl/views/StationLandingPage.scala.html
+++ b/src/main/twirl/views/StationLandingPage.scala.html
@@ -176,6 +176,20 @@
 	}
 }
 
+@otcSpecifics = @{
+	Option(station.specificInfo).collect{
+		case otc: OtcStationSpecifics => otc
+	}
+}
+
+@operationalPeriod = @{
+	station.specificInfo match {
+		case sites: SitesStationSpecifics => sites.operationalPeriod
+		case otc: OtcStationSpecifics => otc.operationalPeriod
+		case _ => None
+	}
+}
+
 @ecoSpecifics = @{
 	Option(station.specificInfo).collect{
 		case eco: EcoStationSpecifics => eco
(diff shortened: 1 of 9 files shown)


Jonathan-Schenk (approved):
Looks good to me! I do not have any comment on this.

src/main/twirl/views/StationLandingPage.scala.html
```diff
@@ -176,6 +176,20 @@ <h2 class="fs-3 mt-5">Pictures</h2>
...
+	Option(station.specificInfo).collect{
+		case otc: OtcStationSpecifics => otc
+	}
+}
+
+@operationalPeriod = @{
+	station.specificInfo match {
+		case sites: SitesStationSpecifics => sites.operationalPeriod
+		case otc: OtcStationSpecifics => otc.operationalPeriod
+		case _ => None
```
ggVGc: I think this is a bit of a code-smell, where we set up a future potential issue where another `StationSpecifics` gets an `operationalPeriod`, everything builds fine, but it is not presented on this page. My first thought is that we should not have the catch-all clause, but after a quick try, it seems I did not get an exhaustiveness error if I leave out the catchall clause, possibly because it's a `.scala.html` file?

However, in this case, since `operationalPeriod` is a `Maybe` anyway, I think we could just make it part of the `StationSpecifics` trait?

ggVGc (approved):
Seems fine, just a small comment, which I think applies to several other similar cases in the code-base, so maybe something to look at as a future improvement (improving the access to `StationSpecifics` data).

src/main/twirl/views/StationLandingPage.scala.html
jonathanthiry: I usually test that I get the result that I want on the page, but I see your point that it could be better to add `operationalPeriod` to the trait and be sure that the page always work as intended.

Removing the catch-all case would throw a runtime exception for other station types. This would throw a warning but, for some reason, [we have disabled them for HTML files](https://github.com/ICOS-Carbon-Portal/meta/blob/64974df6731b0202b4972b429dc6704fd751d5d5/build.sbt?plain=1#L115). I'll create a task to see if we could enable them and fix the warnings.

PR #369, opened 2026-06-05
Add deep links to metadata editor

jonathanthiry:
This updates the URL when navigating in the [metadata editor](https://meta.icos-cp.eu/edit/cpmeta/).
For example https://meta.icos-cp.eu/edit/cpmeta/CentralFacility/CP would take you directly to Carbon Portal under Central Facility.

### DIFF
--- a/src/main/js/metaentry/AppLayout.jsx
+++ b/src/main/js/metaentry/AppLayout.jsx
@@ -8,9 +8,12 @@ var actions = Reflux.createActions([
 ]);
 
 var Backend = require('./backend.js');
+var urlManager = require('./urlManager.js');
 
-var TypesStore = require('./stores/TypesStoreFactory.js')(Backend, actions.selectMetaType, actions.checkUriOrSuffix);
-var IndividualsStore = require('./stores/IndividualsStoreFactory.js')(Backend, actions.selectMetaType, actions.selectIndividual, actions.createIndividual, actions.removeIndividual);
+var initialPath = urlManager.readPath();
+
+var TypesStore = require('./stores/TypesStoreFactory.js')(Backend, actions.selectMetaType, actions.checkUriOrSuffix, initialPath.typeName);
+var IndividualsStore = require('./stores/IndividualsStoreFactory.js')(Backend, actions.selectMetaType, actions.selectIndividual, actions.createIndividual, actions.removeIndividual, initialPath.individualName);
 var EditStore = require('./stores/EditStoreFactory.js')(Backend, actions.selectIndividual, actions.requestUpdate);
 
 var TypesList = require('./views/TypesListFactory.jsx')(TypesStore, actions.selectMetaType);
@@ -19,8 +22,24 @@ var IndividualsList = require('./views/IndividualsListFactory.jsx')(IndividualsS
 var EditView = require('./views/EditViewFactory.jsx')(EditStore, actions.requestUpdate);
 
 module.exports = React.createClass({
-	render: function(){
+	mixins: [Reflux.ListenerMixin],
 
+	getInitialState: function(){
+		return {selectedType: null, selectedIndividual: null};
+	},
+
+	componentDidMount: function(){
+		this.listenTo(TypesStore, function(s){
+			this.setState({selectedType: s.selected});
+			urlManager.updatePath(s.selected, this.state.selectedIndividual);
+		}.bind(this));
+		this.listenTo(IndividualsStore, function(s){
+			this.setState({selectedIndividual: s.selectedIndividual});
+			urlManager.updatePath(this.state.selectedType, s.selectedIndividual);
+		}.bind(this));
+	},
+
+	render: function(){
 		return <div className="row" style={{marginTop: "2px"}}>
 			<div className="col-md-2"><TypesList /></div>
 			<div className="col-md-3"><IndividualsList /></div>
--- a/src/main/js/metaentry/AppLayout.jsx
+++ b/src/main/js/metaentry/AppLayout.jsx
@@ -31,7 +31,7 @@ module.exports = React.createClass({
 	componentDidMount: function(){
 		this.listenTo(TypesStore, function(s){
 			this.setState({selectedType: s.selected});
-			urlManager.updatePath(s.selected, this.state.selectedIndividual);
+			urlManager.updatePath(s.selected, null);
 		}.bind(this));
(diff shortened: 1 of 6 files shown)


src/main/js/metaentry/urlManager.js
```diff
@@ -0,0 +1,37 @@
...
+	};
+}
+
+function updatePath(typeUri, individualUri) {
+	var parts = window.location.pathname.split('/').filter(Boolean);
+	var ontId = parts[1];
+	var path = '/edit/' + ontId + '/';
+	if (typeUri) path += encodeURIComponent(pathName(typeUri));
+	if (typeUri && individualUri) path += '/' + encodeURIComponent(pathName(individualUri));
+	history.pushState(null, '', path);
```
ahd44: There's some issues with this approach, unfortunately.

1) Using `pushState` keeps a trail in history, which is good; but we are not currently listening for if the state changes (a `popState` event), so it's useless in practice. We can add this functionality in but it could end up being complex, hard to tell without doing it.
2) Could replace `pushState` with `replaceState`, which keeps the current URL fresh, and avoids the issue of needing to add functionality. But if someone accidentally clicks away and wants to go back, they can't use the back arrow. (I don't know if anyone **would** try that, though, actually.)

src/main/js/metaentry/urlManager.js
```diff
@@ -0,0 +1,37 @@
...
+
+function pathNameFromLocalName(name) {
+	return name
+		.replace(/([a-z0-9])([A-Z])/g, '$1-$2')
+		.replace(/([A-Z]+)([A-Z][a-z])/g, '$1-$2')
+		.replace(/[_\s]+/g, '-')
+		.replace(/-+/g, '-')
+		.replace(/^-|-$/g, '')
+		.toLowerCase();
+}
```
ahd44: I would suggest removing the path name conversion and using the local name directly.

We already use the local name in the resource URLs ( https://meta.icos-cp.eu/resources/cpmeta/otc_AP_QCflag ) and I think it could be unpredictable for the editing URL to be different ( https://meta.icos-cp.eu/edit/cpmeta/dataset-column/otc-ap-q-cflag ). If we kept the local names, it would be https://meta.icos-cp.eu/edit/cpmeta/DatasetColumn/otc_AP_QCflag which feels more predictable to me.

There is also the danger of the match not being unique. E.g. if for some reason we had both otc_ap_qcflag and otcApQcflag they would end up making the same slug.

src/main/js/metaentry/AppLayout.jsx
```diff
@@ -19,8 +22,24 @@ var IndividualsList = require('./views/IndividualsListFactory.jsx')(IndividualsS
...
+	componentDidMount: function(){
+		this.listenTo(TypesStore, function(s){
+			this.setState({selectedType: s.selected});
+			urlManager.updatePath(s.selected, null);
+		}.bind(this));
+		this.listenTo(IndividualsStore, function(s){
+			this.setState({selectedIndividual: s.selectedIndividual});
+			urlManager.updatePath(this.state.selectedType, s.selectedIndividual);
+		}.bind(this));
+	},
```
ahd44, line 80: I think this is completely appropriate here and is consistent with the code.

But it's kind of a longer-term task to refactor many of our front-end apps, like `metaentry`: convert it to TypeScript, start using functional components, and update React if necessary. This kind of component class with the lifecycle method is no longer recommended, for example, and could be easier to understand and written a bit more clearly with the newer paradigm, I think.

src/main/js/metaentry/stores/IndividualsStoreFactory.js
```diff
@@ -17,6 +19,7 @@ module.exports = function(Backend, selectTypeAction, selectIndividAction, create
 				addingInstance: false, //needed by the IndividualsList view to hide the IndividualAdder when refreshing
 				selectedIndividual: null
 			};
+			this.pendingIndividualName = initialIndividualName || null;
```
ahd44: I prefer the nullish coalescing operator for this:

```suggestion
			this.pendingIndividualName = initialIndividualName ?? null;
```

ahd44 (changes requested):
In general, this looks good and works well. I think it's a really good change.

Left some specific feedback below about some changes that could be made to improve things. Happy to discuss further as needed.

src/main/js/metaentry/urlManager.js
jonathanthiry: Thanks for pointing this out, I wanted to add this feature and didn't do it all the way. I pushed [a commit](https://github.com/ICOS-Carbon-Portal/meta/pull/369/commits/298d79157fde27af12211b8ab2aacf1411723b7a) that should fix this. I think the feature makes the added complexity is worth it.

src/main/js/metaentry/urlManager.js
jonathanthiry: Good point, I changed this in [this commit](https://github.com/ICOS-Carbon-Portal/meta/pull/369/commits/45456a7a8f70118dbfc0ca153aa0ce5f84b59139).

src/main/js/metaentry/AppLayout.jsx
jonathanthiry: I'll leave it as is for now, but we should work on modernizing these apps.

src/main/js/metaentry/stores/IndividualsStoreFactory.js
ahd44: We're going to have to revert this; the version of `babelify` we currently use in the build process uses Babel 5, support for the nullish colaescing operator only arrived in Babel 7. Should be considered when revisiting the code later, but might require larger changes, so best to stick with `|| null` for now.

Sorry about suggesting an incompatible change. The `??` operator was introduced in ES2020 so it's pretty widely available now, but I didn't think about the fact that it might not be included here.

src/main/js/metaentry/urlManager.js
ahd44: It works well in general. One edge case: switching between types of resource breaks the history chain going forward when you go back.  E.g. if you click on Resource A, then click on a different type, then Resource B, if you go back it will correctly load Resource A but incorrectly update the browser state, which then prevents going forward to get to Resource B. It works perfectly fine within the same type, though, and I do think the utility is worth the complexity.

ahd44 (approved):

PR #371, opened 2026-06-18
Improve SPARQL timeout handling and query logging

jonathanthiry:
These changes improve SPARQL timeout handling to reduce the risk of corrupting RDF4J/LMDB transaction state when long-running queries are cancelled:

 - Use RDF4J’s own query timeout via `setMaxExecutionTime`.
 - Avoid immediately force-closing RDF4J query results at timeout.
 - Add a grace-period fallback close if RDF4J timeout does not stop the query.                                                                                   
 - Log SPARQL queries with query id, client IP, hash, response type, query length, time in a separate log file. This would replace the current SPARQL query log done by nginx.

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -12,11 +12,25 @@
 		<minimumBreadcrumbLevel>INFO</minimumBreadcrumbLevel>
 	</appender>
 
+	<appender name="SPARQL_QUERY" class="ch.qos.logback.core.rolling.RollingFileAppender">
+		<file>${SPARQL_QUERY_LOG_FILE:-sparql-query.log}</file>
+		<rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+			<fileNamePattern>${SPARQL_QUERY_LOG_FILE:-sparql-query.log}.%d{yyyy-MM-dd}.gz</fileNamePattern>
+			<maxHistory>14</maxHistory>
+		</rollingPolicy>
+		<encoder>
+			<pattern>%date{ISO8601} %-5level [%logger] [%thread] - %msg%n</pattern>
+		</encoder>
+	</appender>
+
 	<root level="info">
 		<appender-ref ref="STDOUT" />
 		<appender-ref ref="SENTRY" />
 	</root>
 
 	<logger name="org.semanticweb.owlapi.rdf.rdfxml.parser.OWLRDFConsumer" level="WARN" />
+	<logger name="se.lu.nateko.cp.meta.sparql.queries" level="INFO" additivity="false">
+		<appender-ref ref="SPARQL_QUERY" />
+	</logger>
 
 </configuration>
--- a/src/main/scala/services/sparql/Rdf4jSparqlServer.scala
+++ b/src/main/scala/services/sparql/Rdf4jSparqlServer.scala
@@ -110,7 +110,6 @@ class Rdf4jSparqlServer(
 			val queryHash = shortHash(queryStr.query)
 			val errPromise = Promise[ByteString]()
 			val finalizer = new QueryRunFinalizer(qquoter)
-			val permittedLongRunning = new AtomicBoolean(false)
 			log.info(s"SPARQL query started qid=${qquoter.qid} client=${qquoter.cid} hash=$queryHash responseType=${protocolOption.responseType}")
 
 			val sparqlEntityBytes: Source[ByteString, NotUsed] = StreamConverters.asOutputStream(streamTimeout).mapMaterializedValue{ outStr =>
@@ -119,8 +118,7 @@ class Rdf4jSparqlServer(
 				val connCloser = new LoggedCloseOnce(s"SPARQL repository connection qid=${qquoter.qid}", conn, log.debug)
 				val streamCloser = new LoggedCloseOnce(s"SPARQL output stream qid=${qquoter.qid}", new AutoCloseable:
 					def close(): Unit =
-						outStr.flush()
-						outStr.close()
+						try outStr.flush() finally outStr.close()
 				, log.debug)
 
 				val (resultCloser, sparqlFut) = Try:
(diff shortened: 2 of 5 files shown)


src/main/resources/logback.xml
```diff
@@ -12,11 +12,25 @@
...
+	<appender name="SPARQL_QUERY" class="ch.qos.logback.core.rolling.RollingFileAppender">
+		<file>${SPARQL_QUERY_LOG_FILE:-sparql-query.log}</file>
+		<rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+			<fileNamePattern>${SPARQL_QUERY_LOG_FILE:-sparql-query.log}.%d{yyyy-MM-dd}.gz</fileNamePattern>
+			<maxHistory>14</maxHistory>
+		</rollingPolicy>
+		<encoder>
+			<pattern>%date{ISO8601} %-5level [%logger] [%thread] - %msg%n</pattern>
+		</encoder>
+	</appender>
```
ggVGc, line 24: It seems my concern in https://github.com/ICOS-Carbon-Portal/infrastructure/pull/94#pullrequestreview-4563518210 is covered here.

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
 				plainResponse(StatusCodes.BadRequest, userErr.getMessage)
 		}
 
+	private def shortHash(s: String): String =
+		val md = MessageDigest.getInstance("SHA-256").digest(s.getBytes(java.nio.charset.StandardCharsets.UTF_8))
+		md.take(8).map("%02x".format(_)).mkString
+
+	private def logSafeQueryText(query: String): String =
```
ggVGc, line 92: Not entirely clear to me what we're making it "safe" for. Terminal output? I feel like there should be some existing escape functionality in the standard library or elsewhere which we could use?

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
+			.replace("\\", "\\\\")
+			.replace("\r", "\\r")
+			.replace("\n", "\\n")
+			.replace("\t", "\\t")
+			.replace("\"", "\\\"")
+
+	private final class LoggedCloseOnce(name: String, closeable: AutoCloseable, logFailure: String => Unit) extends AutoCloseable:
+		private val closed = new AtomicBoolean(false)
+		override def close(): Unit =
+			if closed.compareAndSet(false, true) then
```
ggVGc, line 104: I don't fully understand why we need this atomic check. How does `closeable.close()` handle `.close()` being called twice? If that is a fatal error, then I guess we should find out about it rather than hiding it with this check?

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
+				try closeable.close()
+				catch case NonFatal(err) => logFailure(s"Failed to close $name: ${err.getClass.getName}: ${err.getMessage}")
+
+	private final class QueryRunFinalizer(qquoter: QueryQuotaManager):
+		private val finished = new AtomicBoolean(false)
+		def finish(): Unit =
+			if finished.compareAndSet(false, true) then qquoter.logQueryFinish()
```
ggVGc, line 111: Same thing here. I don't understand why we need this check. Is finish called multiple times during normal operation?

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
+
 	private def getQueryMarshalling[Q <: Query](
```
ggVGc, line 113: This method has now beome quite large, and I think quite hard to read through. There's a lot going on, and I think at this point we should split it up a bit.

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
 		protocolOption: ProtocolOption[Q]
 	): Marshalling[HttpResponse] = Marshalling.WithFixedContentType(
 		protocolOption.requestedResponseType,
 		() => {
-			val timeout = (config.maxQueryRuntimeSec + 1).seconds
+			val streamTimeout = (config.maxQueryRuntimeSec + 1).seconds
 			val qquoter = quoter.getQueryQuotaManager(queryStr.clientId)
+			val queryHash = shortHash(queryStr.query)
+			val startedAtNanos = System.nanoTime()
+			def elapsedMs: Long = (System.nanoTime() - startedAtNanos) / 1000000
```
ggVGc, line 123: I find the `def` here quite hard to spot. On first read, it looks to me like `elapsedMs` would be a value that is always 0. I would prefer at least writing it as `def elapsedMs()` to make it clearer, and preferably moving it to below the `val` declarations.

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
-				val conn = repo.getConnection()
+			val sparqlEntityBytes: Source[ByteString, NotUsed] = StreamConverters.asOutputStream(streamTimeout).mapMaterializedValue{ outStr =>
 
-				val (closer, sparqlFut) = Try:
-						conn.prepareQuery(queryStr.query).asInstanceOf[Q]
+				val conn = repo.getConnection()
+				val connCloser = new LoggedCloseOnce(s"SPARQL repository connection qid=${qquoter.qid}", conn, log.debug)
+				val streamCloser = new LoggedCloseOnce(s"SPARQL output stream qid=${qquoter.qid}", new AutoCloseable:
+					def close(): Unit =
+						try outStr.flush() finally outStr.close()
```
ggVGc, line 134: This `new Autoclosable` pattern appears quite a lot, and at makes things a bit hard to read, at least for me. I think, unless something like it already exists, we could introduce a helper method which just takes a callback and builds an `AutoClosable` to remove some clutter.

src/main/scala/services/sparql/Rdf4jSparqlServer.scala
```diff
@@ -79,63 +85,116 @@ class Rdf4jSparqlServer(
...
-					catch case _: Throwable =>
-						log.debug("SPARQL stream was closed/detached")
-					finally
-						qquoter.logQueryFinish()
-						conn.close()
-
-				closer
+					streamCloser.close()
+					connCloser.close()
+					finalizer.finish()
```
ggVGc, line 179: Could we leverate https://www.scala-lang.org/api/current/scala/util/Using$.html for this, to reduce the amount of manual cleanup?

src/main/scala/utils/rdf4j/package.scala
```diff
@@ -17,8 +19,21 @@ import java.net.URI as JavaUri
...
 import scala.collection.AbstractIterator
 import scala.util.{Try, Using}
+import scala.util.control.NonFatal
 
 
+private val rdf4jLog = LoggerFactory.getLogger("se.lu.nateko.cp.meta.utils.rdf4j")
+
+private def logIfRepositoryProblem(context: String, err: Throwable): Unit =
+	val msg = Option(err.getMessage).getOrElse("")
+	val isLmdbProblem = msg.contains("MDB_") || msg.contains("Permission denied") ||
```
ggVGc, line 29: How do we know that every "Permission denied" is an lmdb problem?

src/main/scala/utils/rdf4j/package.scala
```diff
@@ -53,9 +68,19 @@ extension (uri: Uri)
...
+					logIfRepositoryProblem("iteration.hasNext", err)
+				throw err
 
-		override def next(): T = res.next()
+		override def next(): T =
+			try res.next()
+			catch case NonFatal(err) =>
+				if !err.isInstanceOf[QueryInterruptedException] then
+					logIfRepositoryProblem("iteration.next", err)
+				throw err
```
ggVGc, line 83: What happens for the unmatched cases? I think we would fail from not handling a case, but maybe that's the same as the exception being rethrown?

ggVGc (commented):
A few general comments, but I also think, if possible, that it would be good to have some kind of test showing how the query execution behaviour is changed now, since there are quite a few changes to how we handle error cases and timeouts. It may be hard to create that test case, I guess, and after the `Virtuoso` change, all of this code will disappear, so possibly not worth it. We will find out if this works in production or not quite quickly anyway :)

PR #372, opened 2026-06-23
Allow RDF type updates

jonathanthiry:
Linked to the OTC metadata update, we have quite a few resources that need to change type and this allows it.


jonathanthiry:
I need this change directly and will already merge this PR. We could rework or even remove it later on.

PR #376, opened 2026-07-20
[virtuoso] [experiment] Enrich statements on-the-fly

ggVGc:
This preserves the original behavior of enriching `hasCitationString`, `hasBiblioInfo` and license on the fly, rather than the approach of pre-materializing triples in https://github.com/ICOS-Carbon-Portal/meta/pull/365.
This should preserve original functionality, but removes the possibility of running the SPARQL endpoint directly against Virtuoso.
At a later point, we could split out this logic from `meta` into a smaller service in front of Virtuoso, which would then still allow us to redeploy `meta` without disturbing the SPARQL endpoint.

[StatementsEnricher.scala](https://github.com/ICOS-Carbon-Portal/meta/blob/valter/virtuoso-enriched/src/main/scala/services/sparql/enrichment/StatementsEnricher.scala#L1) is copied verbatim from `master`, with some renames and minor visibility edits.

### DIFF
--- a/src/main/scala/services/sparql/enrichment/EnrichingRepository.scala
+++ b/src/main/scala/services/sparql/enrichment/EnrichingRepository.scala
@@ -1,0 +1,165 @@
+package se.lu.nateko.cp.meta.services.sparql.enrichment
+
+import scala.language.unsafeNulls
+
+import org.eclipse.rdf4j.common.iteration.CloseableIteration
+import org.eclipse.rdf4j.model.{IRI, Resource, Value, ValueFactory}
+import org.eclipse.rdf4j.query.algebra.evaluation.federation.FederatedServiceResolver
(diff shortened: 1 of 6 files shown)


src/main/scala/services/sparql/enrichment/EnrichingRepository.scala
```diff
@@ -0,0 +1,175 @@
...
+			override def meet(pattern: StatementPattern): Unit = {
+				val predicate = pattern.getPredicateVar
+				if (
+					!predicate.hasValue || predicate.getValue.isInstanceOf[IRI] &&
+					computedPredicates.contains(predicate.getValue.asInstanceOf[IRI])
+				) result = true
+			}
+		})
+		result
+	}
```
ggVGc, line 71: Improve this. Specifically see if we can make an early exit when we reach a true case.

ggVGc:
This performs too poorly to be used, because of many separate requests being made to Virtuoso from RDF4J through `getStatements`.
This brings us back to materialization being the only reasonable choice currently.

